PR #83, opened 2017-04-26
XW-3252 | featured video on scroll behaviour

weronika:
https://wikia-inc.atlassian.net/browse/XW-3252


app/components/article-featured-video.js
```diff
@@ -20,6 +20,30 @@ export default Ember.Component.extend(
...
+		 * Manages video transformation on user's scroll action
+		 *
+		 * @returns {void}
+		 */
+		initOnScrollBehaviour() {
+			var prevScroll = 0,
+				$video = $(".video-container"),
+				videoBottomPosition = $video.offset().top + $video.height();
+
+			$(window).scroll(function() {
```
kvas-damian: This function will be called pretty often. Can limit number of executions using debounce or throttle?

app/components/article-featured-video.js
```diff
@@ -20,6 +20,30 @@ export default Ember.Component.extend(
...
+				var currentScroll = $(window).scrollTop();
+
+				if (!$video.hasClass('fixed') && currentScroll >= videoBottomPosition && currentScroll > prevScroll) {
```
kvas-damian: Why do we need to check previous scroll position? Is it possible that this condition is `true` when you scroll up?

BTW it's nice practice to from lightweight conditions, then browsers can avoid executing more complex code when it founds `false`.

app/components/article-featured-video.js
```diff
@@ -20,6 +20,30 @@ export default Ember.Component.extend(
...
+		},
+
+		/**
+		 * Manages video transformation on user's scroll action
+		 *
+		 * @returns {void}
+		 */
+		initOnScrollBehaviour() {
+			var prevScroll = 0,
+				$video = $(".video-container"),
```
kvas-damian: Probably you should use `this.$()`

app/styles/ooyala/_article-featured-video.scss
```diff
@@ -83,6 +82,81 @@
 				content: '\2022';
 				padding: 0 2px;
 			}
+
+			.video-title {
+				font-size: $wds-typescale-size-base;
+				line-height: 19px;
```
kvas-damian: Can we use line-height that is defined in DS?

app/styles/ooyala/_article-featured-video.scss
```diff
@@ -83,6 +82,81 @@
...
+
+		&.fixed {
+			background: $wds-color-white;
+			box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.2);
+			padding: 12px;
+			position: fixed;
+			top: $wds-global-navigation-height;
+			z-index: $z-site-head + 1;
+			align-items: center;
+			display: flex;
```
kvas-damian: alphabetical order

app/templates/components/article-featured-video.hbs
```diff
@@ -1,6 +1,6 @@
-<div class="video-container{{unless isPlayerLoading ' player-ready'}}"
-	 style="background-image: linear-gradient(to top, rgba(0, 0, 0, 0.9) 0%, transparent 100%), url({{model.url}});">
+<div class="video-container{{unless isPlayerLoading ' player-ready'}}">
 	{{wds-spinner active=isPlayerLoading overlay=false radius=27 strokeWidth=5}}
+	<img src="{{model.url}}" class="video-thumbnail">
```
kvas-damian: :+1:

app/components/article-featured-video.js
rogatty: Also, remember to remove the listener after this component is destroyed.

rogatty:
Please consider using requestAnimationFrame. See:
- http://www.html5rocks.com/en/tutorials/speed/animations/
- https://github.com/WickyNilliams/headroom.js/blob/master/src/Debouncer.js
- https://github.com/WickyNilliams/headroom.js/blob/master/src/Headroom.js#L82

app/components/article-featured-video.js
```diff
@@ -24,6 +24,46 @@ export default Component.extend(
...
+		 *
+		 * @returns {void}
+		 */
+		initOnScrollBehaviour() {
+			let $video = this.$('.video-container'),
+				$siteHead = $('.site-head'),
+				videoBottomPosition = $video.offset().top + $video.height(),
+				showVideoOnScroll = true;
+
+			$(window).on('scroll', {player: this.player, $window: this.$(window)}, function ({data: {player, $window}}) {
```
kvas-damian: Do we need to pass those parameters on every callback? Even when they're not use because throttle decides is not required?

app/components/article-featured-video.js
```diff
@@ -24,6 +24,46 @@ export default Component.extend(
...
+				run.throttle(this, function () {
+					let currentScroll = $window.scrollTop();
+
+					if (currentScroll >= videoBottomPosition && !$video.hasClass('fixed')) {
+						if (showVideoOnScroll && (player === undefined || ! player.isPlaying())) {
+							$video.addClass('fixed');
+							$siteHead.addClass('no-shadow');
```
kvas-damian: Ember suggests using data binding instead direct DOM operations: https://guides.emberjs.com/v2.12.0/templates/conditionals/

app/components/article-featured-video.js
```diff
@@ -24,6 +24,46 @@ export default Component.extend(
...
+						}
+					} else if (currentScroll < videoBottomPosition - $video.height() && $video.hasClass('fixed')) {
+						$video.removeClass('fixed');
+						$siteHead.removeClass('no-shadow');
+					}
+				}, 200);
+			});
+
+			$video.find('.video-close-button').on('click', function () {
```
kvas-damian: Please covert this function to Ember action: https://guides.emberjs.com/v2.12.0/templates/actions/

app/components/article-featured-video.js
```diff
@@ -24,6 +24,46 @@ export default Component.extend(
...
+				$video.removeClass('fixed');
+				showVideoOnScroll = false;
+			});
+
+			$video.find('.video-thumbnail, .video-placeholder').on('click', {player: this.player},
```
kvas-damian: Please covert this function to Ember action: https://guides.emberjs.com/v2.12.0/templates/actions/

app/styles/ooyala/_article-featured-video.scss
```diff
@@ -1,28 +1,128 @@
...
+			padding: 12px;
+			position: fixed;
+			transform: translateY(-100%);
+			top: $wds-global-navigation-height;
+			width: 100%;
+			z-index: $z-site-head - 1;
+
+			.wds-spinner {
+				display: none;
```
kvas-damian: why do we have to hide spinner here? Maybe condition is wrong and it shouldn't be active in this state?

app/styles/ooyala/_article-featured-video.scss
```diff
@@ -1,28 +1,128 @@
...
+				box-shadow: none;
+				border-width: 3px;
+				height: 30px;
+				left: 64px;
+				position: absolute;
+				right: auto;
+				width: 30px;
+
+				.wds-icon {
+					height: 12px;
```
kvas-damian: `.wds-icon-tiny`

app/components/article-featured-video.js
kvas-damian: BTW `site-head` isn't a part of this component. Modifying it's attributes breaks encapsulation. To achieve the goal you should send an action up to the first common parent and then pass data to `site-head`. It's called Data Down Action Up approach in Ember: https://dockyard.com/blog/2015/10/14/best-practices-data-down-actions-up

app/components/article-featured-video.js
```diff
@@ -33,16 +33,16 @@ export default Component.extend(
...
 		 */
 		initOnScrollBehaviour() {
-			var $video = this.$('.video-container'),
+			let $video = this.$('.video-container'),
 				$siteHead = $('.site-head'),
 				videoBottomPosition = $video.offset().top + $video.height(),
 				showVideoOnScroll = true;
 
-			$(window).on('scroll', { player: this.player, $window: this.$(window) }, function({data: {player, $window}}) {
+			$(window).on('scroll', {player: this.player, $window: this.$(window)}, function ({data: {player, $window}}) {
```
kvas-damian: As @rogatty mentioned before. Please remember about unbinding this event when video player is destroyed. Currently it'll stay active on the next pageview and can have unexpected consequences when there is no video.

app/components/article-featured-video.js
```diff
@@ -32,6 +62,8 @@ export default Component.extend(
 			if (this.player) {
 				this.player.destroy();
 			}
+
+			this.$(window).unbind('scroll');
```
rogatty: This will unbind _all_ on-scroll events. We need to name the function passed to `bind` and use it here as the second param.
Also `bind` and `unbind` are deprecated, see https://api.jquery.com/unbind/

app/templates/components/article-featured-video.hbs
```diff
@@ -1,17 +1,28 @@
...
+<div class="video-container">
+	{{! The 'video-wrapper' element is needed here to fix an issue with background color while
+	scrolling on iOS 9. }}
+	<div class="video-wrapper">
+		{{wds-spinner active=isPlayerLoading overlay=false radius=27 strokeWidth=5}}
+		<img src="{{model.url}}" class="video-thumbnail" {{action 'playVideo'}}>
+		<div {{action 'playVideo'}} class="video-placeholder">
+			<div class="video-details">
+				<div class="video-label">
+					Watch
```
kvas-damian: Can we translate this label?

app/components/article-featured-video.js
```diff
@@ -4,11 +4,13 @@ import duration from '../helpers/duration';
 import {track, trackActions} from '../utils/track';
 import extend from '../utils/extend';
 
-const {Component, inject} = Ember;
+const {Component, inject, run} = Ember;
 
 export default Component.extend(
 	{
 		classNames: ['article-featured-video'],
+		classNameBindings: ['isPlayerLoading::player-ready', 'isPlayed:is-played', 'isVideoDrawerVisible:fixed'],
```
hakubo: if isPlayed is bool you can ommit is-played
https://ember-twiddle.com/b5722f90744715f88ae59d2ea92ab5e3?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

it's standard in Ember

app/components/article-featured-video.js
```diff
@@ -23,7 +25,37 @@ export default Component.extend(
...
+						this.set('isVideoDrawerVisible', true);
+						this.toggleSiteHeadShadow(false);
+					} else if (currentScroll < videoBottomPosition - $video.height()) {
+						this.set('videoDrawerDismissed', false);
+						this.closeVideoDrawer();
+					}
+				}, 200);
+			});
+
+			this.$(window).on('scroll', this.get('onScrollHandler'));
```
hakubo: what do you think of using a namespaced event scroll.featured-video
This way you don't have to pass function to off function

https://api.jquery.com/event.namespace/

app/styles/component/_article-featured-video.scss
```diff
@@ -1,30 +1,141 @@
...
+				bottom: 18px;
+				box-shadow: none;
+				border-width: 3px;
+				height: 30px;
+				left: 64px;
+				position: absolute;
+				right: auto;
+				width: 30px;
+
+				.wds-icon-tiny {
```
hakubo: 5 levels deep nesting, what do you think of flattening this file?

app/templates/components/article-featured-video.hbs
latata: As all videos are currently in english so we want "Watch" label everywhere. If we'd extract it to i18n files, we had to put the same translation in every language.

app/styles/component/_site-head.scss
```diff
@@ -11,6 +11,11 @@ $site-head-default-box-shadow: 0 1px 4px 0 $site-head-box-shadow-color;
 
 .site-head-headroom-un-pinned .site-head {
 	box-shadow: $site-head-default-box-shadow;
+
+	&.no-shadow {
+		box-shadow: 0 0 0 0 $site-head-box-shadow-color;
+		transition: box-shadow .3s;
```
hakubo: animiating a box-shadow is slow, is it needed here?

or can we optimize it?
http://tobiasahlin.com/blog/how-to-animate-box-shadow/

app/templates/components/article-featured-video.hbs
rogatty: It's still good practice to have all labels in i18n, even if not translated yet.

app/templates/components/article-featured-video.hbs
```diff
@@ -1,17 +1,28 @@
...
+					<span class="video-time">{{videoTime}}</span>
+				</div>
+				<div class="video-title">{{videoTitle}}</div>
+			</div>
+			<div class="video-play-circle">
+				{{#if isVideoDrawerVisible}}
+					{{svg 'wds-icons-play-triangle' class='wds-icon wds-icon-tiny'}}
```
hakubo: other way to do it would be
```hbs
{{svg 'name' class=(concat 'wds-icon wds-icon-' (if isVideoDrawerVisible 'tiny' 'small'))}}
```

not sure if that's more readable though, I remember that blocked if was 'slow', not sure if that is still a problem

app/components/article-featured-video.js
rogatty: Make sure it works as expected when going from article to article which both have the featured video. Actually, we should check it for the current solution too. Testing notes?

app/templates/components/article-featured-video.hbs
rogatty: For reference:
https://medium.com/square-corner-blog/optimizing-ember-templates-c479d26fe58e#5dc3

But it was for Glimmer 1 and we use Glimmer 2. I don't know which one is better now and have no opinion, sorry :smile:

rogatty: Ok, see https://medium.com/@chancancode/thanks-for-the-analysis-and-i-really-appreciate-that-square-is-sharing-their-work-with-other-ember-aaf004b0aae9 and use what is more readable. I think the original version is easier to understand and grep.

>In Glimmer 2, child blocks are lazily compiled at runtime, so the unused blocks are effectively free (you only pay the cost when you eventually enter them).

app/components/article-featured-video.js
```diff
@@ -23,7 +25,37 @@ export default Component.extend(
 		didRender() {
 			this._super(...arguments);
 
-			this.initVideoPlayer();
+			if (!this.get('hasRendered')) {
```
rogatty: This is misleading. How could it `didRender` but not `hasRendered`? Is the purpose to init the video player only on the first render? If so, the `didInsertElement` hook is called only once, see https://guides.emberjs.com/v2.12.0/components/the-component-lifecycle/#toc_integrating-with-third-party-libraries-with-code-didinsertelement-code

app/components/article-featured-video.js
```diff
@@ -23,7 +25,35 @@ export default Component.extend(
...
+				videoBottomPosition = $video.offset().top + $video.height();
+
+			this.$(window).on('scroll.featured-video', () => {
+				run.throttle(this, () => {
+					let currentScroll = this.$(window).scrollTop();
+
+					if (currentScroll >= videoBottomPosition && this.canVideoDrawerShow()) {
+						this.set('isVideoDrawerVisible', true);
+						this.toggleSiteHeadShadow(false);
+					} else if (currentScroll < videoBottomPosition - $video.height()) {
```
rogatty: `$video.height()` checks the DOM state which is kinda expensive. Can we have `videoTopPosition = $video.offset().top` defined together with `videoBottomPosition` or is the height dynamic in this case?

rogatty: And if it's dynamic and can't be moved, please add a comment explaining that.

rogatty:
I believe this PR should be targeted at `dev`

app/components/article-featured-video.js
```diff
@@ -154,12 +186,48 @@ export default Component.extend(
...
+							label: 'on-scroll-bar'
+						});
+					}
+
 					this.set('isPlayed', true);
+					this.closeVideoDrawer();
 					this.player.play();
 				}
+			},
+			closeVideoDrawer() {
```
rogatty: We have `this.closeVideoDrawer` and `this.actions.closeVideoDrawer`. I'm pretty sure someone at some point will mix them up. Can we distinct them somehow?

app/templates/components/article-wrapper.hbs
```diff
@@ -13,7 +13,8 @@
 {{/if}}
 <section class="article-body">
 	{{#if featuredVideoVisible}}
-		{{article-featured-video model=model.featuredVideo}}
+		{{article-featured-video model=model.featuredVideo
+								 toggleSiteHeadShadow=(action 'toggleSiteHeadShadow')}}
```
rogatty: That's a strange formatting. Let's stick to the convention:
```
{{article-featured-video
	model=model.featuredVideo
	toggleSiteHeadShadow=(action 'toggleSiteHeadShadow')
}}
```

app/controllers/application.js
```diff
@@ -213,6 +213,10 @@ export default Controller.extend(
 			 */
 			toggleUserMenu(visible) {
 				this.set('userMenuVisible', visible);
+			},
+
+			toggleSiteHeadShadow(visible) {
+				this.set('siteHeadNoShadow', !visible);
```
rogatty: I think this negation makes the logic harder to follow and is unnecessary. How about `this.set('siteHeadShadow', visible);` here and `<nav class="site-head{{unless shadow ' no-shadow'}}">` in `site-head.hbs`?

app/components/article-featured-video.js
```diff
@@ -9,8 +9,9 @@ const {Component, inject, run} = Ember;
 export default Component.extend(
 	{
 		classNames: ['article-featured-video'],
-		classNameBindings: ['isPlayerLoading::player-ready', 'isPlayed', 'isVideoDrawerVisible:fixed'],
+		classNameBindings: ['isPlayerLoading::is-player-ready', 'isPlayed', 'isVideoDrawerVisible:is-fixed'],
 		hasRendered: false,
```
rogatty: It's no longer used.

app/components/article-featured-video.js
```diff
@@ -38,19 +35,21 @@ export default Component.extend(
 		 * @returns {void}
 		 */
 		initOnScrollBehaviour() {
-			let $video = this.$('.video-container'),
-				videoBottomPosition = $video.offset().top + $video.height();
+			const $video = this.$('.article-featured-video__container'),
+				videoHeight = $video.height(),
+				videoBottomPosition = $video.offset().top + videoHeight,
+				videoTopPosition = videoBottomPosition - videoHeight;
```
rogatty: Uhmm... We add to substract :see_no_evil: 
```
videoTopPosition = $video.offset().top,
videoBottomPosition = videoTopPosition + videoHeight;
```

app/components/article-featured-video.js
```diff
@@ -60,6 +60,10 @@ export default Component.extend(
 			}
 		},
 
+		getMainElement() {
+			return document.getElementsByClassName(this.get('classNames')[0])[0];
```
latata: `this.$()[0];` instead of `document.getElementsByClassName(this.get('classNames')[0])[0]`

PR #85, opened 2017-04-28
XW-3310 | Refactor Models to be able to inject owner

kvas-damian:
## Links

* http://wikia-inc.atlassian.net/browse/XW-3310

## Description

Recently we had a problem with [`getService()` method which was leaking `application` between requests](https://github.com/Wikia/mobile-wiki/pull/84). To allow using services in Models we decided to refactor them to start using Object instances instead of static methods. This approach allows us to inject `owner` into `EmberObject` instance. `EmberObject` with owner injected can easily access to Ember services.

**It's recommended to create new models using**: `Model.create(getOwner(this).ownerInjection(), originalParams);`

## Reviewers

@Wikia/x-wing @Wikia/adeng @slayful


app/components/article-comments.js
```diff
@@ -85,7 +85,7 @@ export default Component.extend(
 		init() {
 			this._super(...arguments);
 
-			this.set('model', ArticleCommentsModel.create({
+			this.set('model', ArticleCommentsModel.create(getOwner(this).ownerInjection(), {
```
latata, line 88: Maybe we could create our custom static method for creating models? Something like `ArticleCommentsModel.createModel(this, {data});` or helper method for `getOwner(this).ownerInjection()`

app/models/article-edit.js
```diff
@@ -3,102 +3,108 @@ import getEditToken from '../utils/edit-token';
...
 
-const ArticleEditModel = Ember.Object.extend({
+const {
+	Object: EmberObject,
+	computed,
+	get,
+	inject
+} = Ember;
+
+export default EmberObject.extend({
```
latata: `EmberObject` or `Object`?

app/components/article-comments.js
kvas-damian: I believe passing current object instance into a constructor breaks SOLID principle.
Helper method sounds like a valid alternative. Do you think it will simplify the code? We'd need to import that helper..

app/mixins/wiki-page-handler.js
```diff
@@ -0,0 +1,143 @@
+import Ember from 'ember';
+import {defineError} from 'ember-exex/error';
```
rogatty: It's not used.

app/models/article-edit.js
```diff
@@ -3,102 +3,108 @@ import getEditToken from '../utils/edit-token';
...
-					action: 'query',
-					prop: 'revisions',
-					// FIXME: It should be possible to pass props as an array
-					rvprop: 'content|timestamp',
-					titles: title,
-					rvsection: sectionIndex,
-					format: 'json',
-					cb: new Date().getTime()
+	/**
+	 * @param {string} host
```
rogatty: It's no longer valid.

app/routes/wiki-page.js
```diff
@@ -99,9 +99,8 @@ export default Route.extend(
 				modelParams.page = params.page;
 			}
 
-			return RSVP.resolve(getPageModel(
+			return RSVP.resolve(this.getPageModel(
 				modelParams,
-				fastboot,
 				this.get('wikiVariables.contentNamespaces')
```
rogatty: There is no need to pass this down. But maybe it's easier to reason about if we pass it from here? I'm not sure.

rogatty (approved):

rybmat (approved):

PR #86, opened 2017-05-02
XW-3206 | Correct response status codes in Express

kvas-damian:
## Links

* http://wikia-inc.atlassian.net/browse/XW-3206

## Description

Previously we were using incorrect statusCode that we finally served to client. In this PR on error we get highest statusCode and serve it to browser.

## Reviewers

@rogatty

### DIFF
--- a/server/app/server.js
+++ b/server/app/server.js
@@ -55,7 +55,7 @@ const server = new FastBootAppServer({
 		app.use((err, req, res, next) => {
 			if (err) {
 				// Handle errors that don't go to FastBoot, like Bad Request etc.
-				const statusCode = Math.max(res.statusCode, err.statusCode);
+				const statusCode = Math.max(res.statusCode, err.statusCode || 500);
 				const level = levelFn(statusCode);
 				const logFn = req.log[level].bind(req.log);
 
--- a/server/app/server.js
+++ b/server/app/server.js
@@ -19,14 +19,15 @@ const staticAssets = require('../static-assets');
 const methodOverride = require('method-override');
 const bodyParser = require('body-parser');
 
-function levelFn(status, err) {
-	if (err || status >= 500) {
+function levelFn(status) {
+	if (status >= 500) {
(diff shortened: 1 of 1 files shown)


server/app/server.js
```diff
@@ -53,14 +54,14 @@ const server = new FastBootAppServer({
 	afterMiddleware: (app) => {
 		app.use((err, req, res, next) => {
 			if (err) {
-				// It should never get here
-				// In theory, the Ember app handles all server errors
-				const level = levelFn(res.statusCode, err);
+				// Handle errors that don't go to FastBoot, like Bad Request etc.
+				const statusCode = Math.max(res.statusCode, err.statusCode);
```
rogatty: Can we have a situation where there is no `res.statusCode` or `err.statusCode`?

```
> var x = {};
> Math.max(10, x.y)
< NaN
```

server/app/server.js
```diff
@@ -53,14 +54,14 @@ const server = new FastBootAppServer({
...
+				const level = levelFn(statusCode);
 				const logFn = req.log[level].bind(req.log);
 
 				logFn(err);
 
-				res.send('Server error');
+				res.sendStatus(statusCode);
```
rogatty, line 64: Does it send a blank page now or a stack trace?

kvas-damian: are you referring to Ember or express stack trace?

`sendStatus()` sends given status code with default message:
https://expressjs.com/en/api.html#res.sendStatus

rogatty: Oh, the default message is :+1:

server/app/server.js
kvas-damian: I didn't found such case. Do you think we fallback to `500` in such cache?

rogatty: How about `Math.max(parseInt(res.statusCode, 10), parseInt(err.statusCode, 10)) || 500`?

rogatty (approved):

PR #87, opened 2017-05-02
XW-3241 | Don't redirect client when api returns 200 without redirect

kvas-damian:
## Links

* http://wikia-inc.atlassian.net/browse/XW-3241

## Description

When API returned 200 with HTML response we're redirecting browser to API url. It's correct behavior only when API returns redirect and then we get HTML. In this case we should server error.

## Reviewers

@rogatty

### DIFF
--- a/app/models/wiki-variables.js
+++ b/app/models/wiki-variables.js
@@ -35,9 +35,21 @@ WikiVariablesModel.reopenClass({
 
 				if (contentType && contentType.indexOf('application/json') !== -1) {
 					return response.json();
-				} else {
+				} else if (url !== response.url) {
+					// API was redirected to non-json page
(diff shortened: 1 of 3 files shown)


app/models/wiki-variables.js
```diff
@@ -35,10 +35,22 @@ WikiVariablesModel.reopenClass({
 
 				if (contentType && contentType.indexOf('application/json') !== -1) {
 					return response.json();
-				} else {
+				} else if (url !== response.url) {
+					// API was redirected to non-json page
 					throw new NonJsonApiResponseError().withAdditionalData({
```
rogatty: This error's name doesn't fit anymore. Below we have `// non-json API response` and it's not used there. How about `WikiVariablesRedirectError`?

rogatty (approved):

PR #92, opened 2017-05-12
XW-3262 | Featured Video on articles with Portable Infoboxes

rybmat:
## Links

https://wikia-inc.atlassian.net/browse/XW-3262

## Description

featured video on articles with portable infoboxes

## Reviewers

@Wikia/x-wing


app/components/article-featured-video.js
```diff
@@ -43,7 +46,7 @@ export default Component.extend(
 		 */
 		initOnScrollBehaviour() {
 			const $video = this.$('.article-featured-video__container'),
-				videoHeight = $video.height(),
+				videoHeight = this.get('withinPortableInfobox') ? 97 : 230,
```
latata: can't we get these values from somewhere, from DOM?

app/components/article-featured-video.js
```diff
@@ -92,6 +95,17 @@ export default Component.extend(
 				});
 			});
 
+			// when playing video on article with infobox, closing fullscreen also has to pause video
+			// as it will be not visible
+			player.mb.subscribe(window.OO.EVENTS.FULLSCREEN_CHANGED, 'ui-display-update', (name, isFullScreen, paused) => {
+				this.set('isPlaying', isFullScreen);
```
latata: we change `isPlaying` event if it's article without infobox? And it's confusing assigning isFullScreen to isPlaying

app/templates/components/article-wrapper.hbs
```diff
@@ -11,14 +11,16 @@
...
 {{/if}}
-<section class="article-body">
-	{{#if featuredVideoVisible}}
-		{{article-featured-video
-				model=model.featuredVideo
-				toggleSiteHeadShadow=(action 'toggleSiteHeadShadow')
-		}}
-	{{/if}}
 
+{{#if hasFeaturedVideo }}
```
latata: remove space before `}}`

tests/unit/components/article-wrapper-test.js
```diff
@@ -112,26 +101,3 @@ test('wiki with disableAnonymousEditing set requires log in', function (assert)
 		expected: false
 	}, 'editAllowed', assert);
 });
-
-test('featured video visible for articles without infoboxes and featured video set', function (assert) {
```
latata, line 116: don't we want to leave a test checking if video is visible for articles with featured video set? or it's too obvious?

hakubo (approved):

app/styles/component/_article-featured-video.scss
```diff
@@ -251,4 +215,86 @@
 			}
 		}
 	}
+
+	&.within-portable-infobox {
+		//make sure it is the same as $infobox-background in _portable-infobox.scss
+		$infobox-bg-color: mix(mix($button-color, $black, 50%), $white, 10%);
```
latata, line 221: can't we extract this variable to be available from both this file and `_portable-infobox.scss`?

app/components/article-featured-video.js
rybmat: it happened few times that I got 0 as a height taken from DOM, and since video height is constant this in my opinion is simplest solution

app/components/article-featured-video.js
rybmat: confusing? play only if it is in fullScreen. And yes, this line should be within if below

app/styles/component/_article-featured-video.scss
rybmat: where do you propose to extract this?

tests/unit/components/article-wrapper-test.js
rybmat: don't think that `hasFeaturedVideo: computed.bool('model.featuredVideo'),` needs a test

app/components/article-featured-video.js
rogatty: The video height isn't constant, it depends on the viewport width, we keep the aspect ratio.

app/components/article-featured-video.js
```diff
@@ -18,23 +18,27 @@ let lastTimestamp = 0,
 export default Component.extend(
 	{
 		classNames: ['article-featured-video'],
-		classNameBindings: ['isPlayerLoading::is-player-ready', 'isPlayed'],
+		classNameBindings: ['isPlayerLoading::is-player-ready', 'hasStartedPlaying', 'isPlaying',
+							'withinPortableInfobox:within-portable-infobox:without-portable-infobox'],
```
rogatty: Let's format it nicely, binding per line.

hakubo (approved):

hakubo (approved):

hakubo (approved):

PR #93, opened 2017-05-15
XW-3356 | Fix for memory leak

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/XW-3356

## Description

`utils/application-instance` was shared between FastBoot requests which caused data pollution and possibly the memory leak we see in the metrics. Instead of using this hack, let's use an Ember service.

Additionally, we will be able to turn on [heapdump](https://github.com/bnoordhuis/node-heapdump) using chef and dump the memory heap on production when we see a leak.

## Reviewers

@Wikia/x-wing


app/mixins/wiki-page-handler.js
```diff
@@ -47,9 +46,10 @@ export default Ember.Mixin.create({
 	getPageModel(params) {
 		const isFastBoot = this.get('fastboot.isFastBoot'),
 			shoebox = this.get('fastboot.shoebox'),
-			contentNamespaces = this.get('wikiVariables.contentNamespaces');
+			contentNamespaces = this.get('wikiVariables.contentNamespaces'),
+			isInitialPageView = this.get('initialPageView').isInitialPageView(getOwner(this));
```
hakubo: I don't see isInitialPageView to accept parameters

hakubo (changes requested):

app/mixins/wiki-page-handler.js
rogatty: It was accepting them... for a while. Cleaned up.

hakubo (approved):

PR #101, opened 2017-05-24
Ser-968 In context translations

jcellary:
## Links

* http://wikia-inc.atlassian.net/browse/Ser-968

## Description

This adds support for in context translations which will be optionally run on sandbox-i18n-* machines. Related to https://github.com/Wikia/chef-repo/pull/13737


kvas-damian (approved):

kvas-damian:
Ping @rogatty what do you think?

@jcellary Does crowdin script update `lol` language file as well when downloading/uploading translations?

rogatty (approved):
I'd change the locale name to something like `ict`. `lol` is misleading.
Apart from that LGTM :+1:

jcellary:
The download script downloads all languages that are defined in crowdin, so if we add another language it will be automatically included. I used 'lol', because we need to select a language from one of the predefined codes from a dropdown and lol is not a real language, hence it will not interfere with our real translations.

PR #105, opened 2017-06-07
ADEN-5928 Refactor ads flow and use slotsContext

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-5299

### DIFF
--- a/app/mixins/ads.js
+++ b/app/mixins/ads.js
@@ -38,6 +38,10 @@ export default Ember.Mixin.create({
 			config = adsData.additionalConfig[adSlotName] || {},
 			$placeholder = $('<div>');
 
+		if (!this.get('ads.module').isSlotApplicable(adSlotName)) {
+			return;
+		}
+
 		if (place === 'after') {
 			$placeholder.insertAfter(element);
 		} else if (place === 'before') {
@@ -96,15 +100,15 @@ export default Ember.Mixin.create({
 			this.appendAd(adsData.mobileTopLeaderBoard, 'after', $pageHeader.first());
 		}
 
-		if (this.get('ads.module').isSlotApplicable(adsData.mobileInContent)) {
+		if ($firstSection.length > 0) {
 			this.appendAd(adsData.mobileInContent, 'before', $firstSection);
 		}
 
-		if (this.get('ads.module').isSlotApplicable(adsData.mobilePreFooter)) {
+		if ($articleFooter.length > 0) {
 			this.appendAd(adsData.mobilePreFooter, 'before', $articleFooter);
 		}
 
-		if (this.get('ads.module').isSlotApplicable(adsData.mobileBottomLeaderBoard)) {
+		if ($globalFooter.length > 0) {
 			this.appendAd(adsData.mobileBottomLeaderBoard, 'before', $globalFooter);
 		}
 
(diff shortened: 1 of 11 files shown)


app/mixins/ads.js
```diff
@@ -40,6 +38,10 @@ export default Ember.Mixin.create({
 			config = adsData.additionalConfig[adSlotName] || {},
 			$placeholder = $('<div>');
 
+		if (!this.get('ads.module').isSlotApplicable(adSlotName)) {
```
dianafa: nitpick - what do you think about doing this check before assigning the values to vars above and querying the dom for the first div?

app/modules/ads.js
```diff
@@ -349,6 +365,64 @@ class Ads {
 		}
 	}
 
+	isTopLeaderboardApplicable() {
+		const hasFeaturedVideo = this.getTargetingValue('hasFeaturedVideo'),
+			isHome = this.getTargetingValue('pageType') === 'home',
+			$pi = $('.portable-infobox'),
+			$pageHeader = $('.wiki-page-header');
+
+		return isHome || $pi.length > 0 || ($pageHeader.length > 0 && !hasFeaturedVideo);
```
dianafa: nitpick: what about:
```
hasPortableInfobox = $('.portable-infobox').length > 0,
hasPageHeader = $('.wiki-page-header').length > 0
```
and then in final check it would be:
```
isHome || hasPortableInfobox || (hasPageHeader && !hasFeaturedVideo)
```
??

app/modules/ads.js
```diff
@@ -349,6 +365,64 @@ class Ads {
...
+			$articleFooter = $('.article-footer'),
+			articleBodyHeight = $articleBody.height(),
+			showInContent = this.isInContentApplicable(),
+			$trendingArticles = $('.trending-articles');
+
+		if (this.getTargetingValue('pageType') === 'home') {
+			return $trendingArticles.length > 0;
+		}
+
+		return $articleFooter.length && !showInContent || articleBodyHeight > this.adsData.minPageLength;
```
dianafa: the same question as above, what about: hasArticleFooter = $('.article-footer').length > 0

app/modules/ads.js
```diff
@@ -349,6 +365,64 @@ class Ads {
...
+	}
+
+	isPrefooterApplicable() {
+		const $articleBody = $('.article-body'),
+			$articleFooter = $('.article-footer'),
+			articleBodyHeight = $articleBody.height(),
+			showInContent = this.isInContentApplicable(),
+			$trendingArticles = $('.trending-articles');
+
+		if (this.getTargetingValue('pageType') === 'home') {
```
dianafa:
```
isHome = this.getTargetingValue('pageType') === 'home'
hasTrendingArticles = $('.trending-articles').length > 0
```
and then
```
if (isHome && hasTrendingArticles) {
return true;
}
```
would be more readable, but its a nitpick again :)

app/modules/ads.js
```diff
@@ -365,21 +439,22 @@ class Ads {
...
-				this.mercuryPV = this.mercuryPV + 1;
-				this.adLogicPageParams.add('mercuryPV', this.mercuryPV.toString());
-			});
+			this.setupSlotsContext();
+			if (this.adMercuryListenerModule) {
+				this.adMercuryListenerModule.runOnPageChangeCallbacks();
+				this.adMercuryListenerModule.onPageChange(() => {
+					this.adLogicPageViewCounterModule.increment();
+					this.googleTagModule.updateCorrelator();
+					this.mercuryPV = this.mercuryPV + 1;
```
dianafa: `this.mercuryPV +=1` ? :)

app/modules/ads.js
```diff
@@ -349,6 +365,64 @@ class Ads {
...
+		this.slotsContext.setStatus('MOBILE_PREFOOTER', this.isPrefooterApplicable());
+		this.slotsContext.setStatus('MOBILE_BOTTOM_LEADERBOARD', this.isBottomLeaderboardApplicable());
+	}
+
+	isSlotApplicable(slotName) {
+		if (!this.slotsContext) {
+			return true;
+		}
+
+		return this.slotsContext.isApplicable(slotName);
```
dianafa: what about
```
return !this.slotsContext || this.slotsContext.isApplicable(slotName)
``` 
?

app/modules/ads.js
fraszczakszymon: 1. :+1:
2. :-1: - we want to return false when it's home and page doesn't have trending articles

dianafa:
LGTM

nandy-andy (approved):

PR #107, opened 2017-06-08
(PLATFORM-3071) Send internal request header when calling Helios

Grunny:
## Links

* https://wikia-inc.atlassian.net/browse/PLATFORM-3071

## Description

Send internal request header when calling Helios.

## Reviewers

@rogatty

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -43,6 +43,7 @@ export default EmberObject.extend({
 		});
 
 		return fetch(`${config.helios.internalUrl}${queryString}`, {
+			headers: {'X-Wikia-Internal-Request': '1'},
 			timeout: config.helios.timeout,
 		}).then((response) => {
 			if (response.ok) {


harnash (approved):

rogatty (approved):

app/models/user.js
```diff
@@ -43,6 +43,7 @@ export default EmberObject.extend({
 		});
 
 		return fetch(`${config.helios.internalUrl}${queryString}`, {
+			headers: {'X-Wikia-Internal-Request': '1'},
```
harnash, line 46: I just wonder if we shouldn't use this like this:
```js
const headers = new Headers({'X-Wikia-Internal-Request': '1'});
```

and then supply it to the `fetch()` (See: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)

rogatty: ember-network uses node-fetch on server side, see https://github.com/bitinn/node-fetch#usage for examples. They don't use `Headers` in this context.

harnash: 👍

PR #116, opened 2017-07-03
XW-2830 | Make sure that Local Nav Links Expressed in HTML

JamesHawking:
## Links

* https://wikia-inc.atlassian.net/browse/XW-2830

## Description
Local nav is one of the only elements that is not prerendered on the mobile page, but need to express those critical internal links in HTML to be as safe as possible.
Always render local nav (all levels) and hide them via css

## Reviewers

@Wikia/x-wing


app/components/fastboot-only/local-navigation-for-seo.js
```diff
@@ -0,0 +1,35 @@
...
+	inject
+} = Ember;
+
+export default Component.extend({
+	classNames: ['local-navigation-for-seo'],
+	wikiVariables: inject.service(),
+	localLinks: computed.reads('wikiVariables.localNav'),
+	currentLocalLinks: computed.or('currentLocalNav.children', 'localLinks'),
+	flatNavigationLinks: computed('currentLocalLinks', function () {
+		let deepMap = function (linksList) {
```
hakubo: please extract this method outside component, so it does not have to be created anytime currentLocalLinks is updated.

app/components/fastboot-only/local-navigation-for-seo.js
```diff
@@ -0,0 +1,35 @@
...
+					flatArray = flatArray.concat(deepMap(item.children));
+				}
+			});
+
+			return flatArray;
+		};
+
+		return deepMap(this.get('currentLocalLinks'));
+	}),
+	localNavigationForSeo: computed('flatNavigationLinks', function () {
```
hakubo: computed.reads? or .alias?

hakubo (changes requested):

app/components/fastboot-only/local-navigation-for-seo.js
rogatty: How about calling it `model`? That's what it is.

hakubo (approved):

rogatty (approved):

PR #127, opened 2017-07-21
XW-3423 | when redirecting to oasis append query param instead of overriding all of them

rybmat:
## Links
https://wikia-inc.atlassian.net/browse/XW-3423

## Reviewers

@Wikia/x-wing

### DIFF
--- a/app/routes/wiki-page.js
+++ b/app/routes/wiki-page.js
@@ -150,7 +150,7 @@ export default Route.extend(
 						redirectTo = buildUrl({
 							host: this.get('wikiVariables.host'),
 							wikiPage: get(transition, 'params.wiki-page.title'),
-							query: Object.assign(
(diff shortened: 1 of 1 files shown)


app/routes/wiki-page.js
```diff
@@ -150,9 +150,11 @@ export default Route.extend(
 						redirectTo = buildUrl({
 							host: this.get('wikiVariables.host'),
 							wikiPage: get(transition, 'params.wiki-page.title'),
-							query: {
-								useskin: 'oasis'
-							}
+							query: Object.assign(
```
rogatty: It won't work on every browser. Use `extend()` which is already imported.

rogatty (approved):

hakubo (approved):

PR #129, opened 2017-07-25
XW-3707 | Integrate Plista into mixed content footer in Australia and New Zealand

JamesHawking:
## Links

* https://wikia-inc.atlassian.net/browse/XW-3707

## Description

As part of GroupM in the AU/NZ region, we need to implement one of their native ads via their API into the Impact footer on Desktop and the recirculation footer on Mobile.

## Reviewers
@Wikia/x-wing


app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
 		isVisible: false,
 		liftigniter: inject.service(),
 		i18n: inject.service(),
+		hasNoLiftigniterSponsoredItem: true,
+		isInRightCountry: false,
+		shouldShowPlista: computed.and('hasNoLiftigniterSponsoredItem', 'isInRightCountry'),
```
hakubo: shouldShowPlista: comupted(function() {
 return ['AU', 'NZ'].indexOf(M.Geo....) > -1 * this.get('hasNo...')
})

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
 		isVisible: false,
 		liftigniter: inject.service(),
 		i18n: inject.service(),
+		hasNoLiftigniterSponsoredItem: true,
```
hakubo: I would reduce it here over items here not to couple this with didEnterViewport and side effect of some other maps

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
...
+		isInRightCountry: false,
+		shouldShowPlista: computed.and('hasNoLiftigniterSponsoredItem', 'isInRightCountry'),
+		plistaSponsoredContent: [],
+		fetchPlista() {
+			const plistaURL = 'http://farm.plista.com/recommendation/?publickey=845c651d11cf72a0f766713f&widgetname=api' +
+							'&count=1&adcount=1&image[width]=583&image[height]=328';
+			return fetch(plistaURL)
+				.then(response => {
```
hakubo: .then((response) => response.json())

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
...
+					return response.json();
+				})
+				.then(data => {
+					if (data.length) {
+						return data[0];
+					} else {
+						throw new Error('We haven\'t got PLISTA!');
+					}
+				})
+				.catch(error => {
```
hakubo: we do not want to log errors to console

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
...
+							'&count=1&adcount=1&image[width]=583&image[height]=328';
+			return fetch(plistaURL)
+				.then(response => {
+					return response.json();
+				})
+				.then(data => {
+					if (data.length) {
+						return data[0];
+					} else {
+						throw new Error('We haven\'t got PLISTA!');
```
hakubo: what does it mean? why we haven't got it?
I'd focus this on "Plista returned no recomendations" or something like that

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
...
+					}
+				})
+				.catch(error => {
+					console.log(error);
+				});
+		},
+		mapPlista(item) {
+			if (typeof(item) !== 'undefined') {
```
hakubo: you throw if plista has not returned data
is this check necessary here?

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,40 @@ export default Component.extend(
...
+				return {
+					meta: 'wikia-impactfooter',
+					thumbnail: item.img,
+					title: item.title,
+					url: item.url,
+					presented_by: item.brand,
+					plista: true
```
hakubo: this is a boolean
what do you think about
isPlista

hakubo (approved):

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,39 @@ export default Component.extend(
 		isVisible: false,
 		liftigniter: inject.service(),
 		i18n: inject.service(),
+		hasNoLiftigniterSponsoredItem: computed('items', function () {
+			return !this.get('items').some((item) => item.presented_by);
+		}),
+		isInRightCountry: false,
```
rogatty: It's not used anywhere.

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,39 @@ export default Component.extend(
...
+		shouldShowPlista: computed('hasNoLiftigniterSponsoredItem', function () {
+			return ['AU', 'NZ'].indexOf(M.geo.country) > -1 && this.get('hasNoLiftigniterSponsoredItem');
+		}),
+		fetchPlista() {
+			const plistaURL = 'http://farm.plista.com/recommendation/?publickey=845c651d11cf72a0f766713f&widgetname=api' +
+							'&count=1&adcount=1&image[width]=583&image[height]=328';
```
rogatty: How about we use the same dimensions as for LiftIgniter images?
```
width: normalizeThumbWidth(window.innerWidth)
```

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,39 @@ export default Component.extend(
...
+				.then((response) => response.json())
+				.then((data) => {
+					if (data.length) {
+						return data[0];
+					} else {
+						throw new Error('We haven\'t got PLISTA!');
+					}
+				})
+				.catch((error) => {
+					// If Plista did not return anything, just don't add it to to items
```
rogatty: What do you think about `Logger.info('Plista fetch failed', error)`? Right now we swallow the errors and nothing is logged.

I wouldn't throw the error when the response is empty, though. It's not an error, it's just targeting, right?

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,39 @@ export default Component.extend(
...
+				});
+		},
+		mapPlista(item) {
+			return {
+				meta: 'wikia-impactfooter',
+				thumbnail: item.img,
```
rogatty: When `fetchPlista()` doesn't return any item (in other words - returns `undefined`), this will throw an error.

app/components/recirculation-prefooter.js
```diff
@@ -56,6 +88,17 @@ export default Component.extend(
 							'LI'
 						);
 					});
+
+					if (this.get('shouldShowPlista')) {
+						this.fetchPlista()
+							.then(this.mapPlista)
+							.then((item) => {
+								if (item.thumbnail) {
+									this.notifyPropertyChange('items');
```
rogatty: I think it should be called below the setter. Even if it doesn't matter and Ember handles it anyway, it will be easier to reason about.

app/components/recirculation-prefooter.js
```diff
@@ -56,6 +88,17 @@ export default Component.extend(
...
+									this.set('items.1', item);
```
rogatty: If Plista is slow to return the data, it's possible we display a Fandom article and then replace it with an ad, right? IMO it's a poor UX, user might like the article, wanted to click it and it disappeared. It would be better to inject ad between the existing items.

app/styles/component/_recirculation.scss
```diff
@@ -30,17 +30,28 @@
...
...IE7ocYjYWPvSwQCnCIy6L4rWPcpR1JbZS//XCoXJZgL38scf5Eq4APbGhwlhYWcXV1GqHPaxmo5i6afOWrbZt2yd55Ulgh4Vm+POUcY61vDzsJDlFquV6HXbtrtvq5Ch7UO9RbJHYmzxvdTlymZ6YAGflUw2oM7hR9jRiX+5mKWU/IzmAiqZgJw6yxzPzkw8dPqLmP+2N8gxg5SnTbJR0KnKe1YJUg6NELeITrXjuGNFRO0c5jhPkiRJ/kmidkswWakN7INGpcDzvRunAC+y6HJ1ELyKdldxph7SUaCe6iJu/5I1C3VF8xPrcy4Ck4hU+qsV8dj0DPza/B5v4oxKOchYJXihn+S0DMUS90g6uvwqzC6UPASI0LeeIKtDH9ddQydPwPHNtVNV4ZmooBlNBEwmny/Q6a3m6fq4fFp6kaFwgspWKORs5ShfCvVBExAcwSFG4qD3Lh1VDKUaw2+4UYhdyKcEllipGNA5diJmwjMsrGyToUzWULIaiFnA1BS5hX7b7CfM3OMp++LlL/vKb8lgFDWCYYuDQXvQiTkPJw1vubw2618NeNnp7aAn7554+b5RZADEUWS9nMhAo7xD/n4udQuwsXk5JNNZ3RQWN4D3X9Lmo8UBZg0btxRVFr6G0urxymSUBijy8KS8+b57gYBfKbo7CingFJEViWwGPxiK8r9lks5Q7gg+4+qFAKUO01veOmVHmgUKrcIbrV3kMBMowMvgD/RWbTCmSKOnaa61E15N4QkTUwzIZxWChvCDecg8ZxC7dHRQbqNl8hP7TzQgwEW0v17qtA2mD7D7cgRGLJ0obOlilmfah2l1BObjBrh8jm2gt2tlLSxovr1GwPKYqo20qrzSHgY1yTK6uEoptbiP+70lwNn1SXqTtQd2fQro0MKLAzLqFBwprHZ91ZtowOtZXrtaj/fN07GK7lfUEtmR+UYQDFt5XdF+23PSTzk2BUsHbm238RNeL426jrxDP0sJhb7jW7NqZLV9/U1ga9u11K3mhiIaan3NjQret+s6mNNwUtWJp+96e7KgWs3748Sd1zkDXeOqMp61OkUp/9uWhheWGiaH8AlXCrx5b4T+PAoxmRf5m7cfvqXrfHnAZigTFyP3h65M2P9uq6CE59c9t/2ap4F5Z72sKIcUIGrLofXzW/1+mL38A8Q75z/jUvY4AAAAASUVORK5CYII=');
```
rogatty: Can we use a separate file? This way, we will load it for **all users** and it will postpone the first render.

app/templates/components/recirculation-prefooter.hbs
```diff
@@ -12,6 +15,12 @@
 								 class="recirculation-prefooter__item-is-video">
 						{{/if}}
 					</div>
+					{{#if post.isPlista}}
+						<a {{action 'adChoiceBadgeClick' bubbles=false}}
+								class="recirculation-prefooter__badge
+							recirculation-prefooter__badge-plista" href="https://www.plista.com/optout/index">
```
rogatty: https://www.plista.com/opt-out is the working URL

app/templates/components/recirculation-prefooter.hbs
```diff
@@ -2,7 +2,10 @@
 	<h3>{{i18n 'recirculation.trending-fandom-articles'}}</h3>
 	<ul class="recirculation-prefooter__list">
 		{{#each items as |post index|}}
-			<li class="recirculation-prefooter__item" {{action 'postClick' post index preventDefault=true}}>
+			<li class="recirculation-prefooter__item" {{action 'postClick'
+				post
+				index
+				preventDefault=true}}>
```
rogatty: I don't think it's more readable this way.

app/components/recirculation-prefooter.js
```diff
@@ -24,6 +24,39 @@ export default Component.extend(
...
+				});
+		},
+		mapPlista(item) {
+			return {
+				meta: 'wikia-impactfooter',
+				thumbnail: item.img,
+				title: item.title,
+				url: item.url,
+				presented_by: item.brand,
+				isPlista: true
```
rogatty: Should we also add `source: 'plista'` as @hakubo did?

app/templates/components/recirculation-prefooter.hbs
```diff
@@ -19,6 +16,7 @@
 						<a {{action 'adChoiceBadgeClick' bubbles=false}}
 								class="recirculation-prefooter__badge
 							recirculation-prefooter__badge-plista" href="https://www.plista.com/opt-out">
+							<img src="https://services.wikia.com/static-assets/image/c7c31771-d3e2-4aee-8d0d-6d4c4629bcff">
```
rogatty: `http://vignette.wikia.nocookie.net/c7c31771-d3e2-4aee-8d0d-6d4c4629bcff` is a proper, public-facing URL. See https://wikia-inc.atlassian.net/wiki/display/PLAT/Static+Assets

PR #150, opened 2017-09-05
ADEN-5812 New outstream look&feel

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-5812

Wait for https://github.com/Wikia/app/pull/13661

### DIFF
--- a/app/styles/module/_ads-porvata.scss
+++ b/app/styles/module/_ads-porvata.scss
@@ -1,0 +1,171 @@
+@import 'ads-mixin';
+
+$uap-progress-bar-color: #b7b7b7;
+$uap-progress-bar-current-time-color: #666;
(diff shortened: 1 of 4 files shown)


app/styles/module/_ads-porvata.scss
```diff
@@ -0,0 +1,177 @@
...
+			width: $player-button-size;
+		}
+	}
+}
+
+.ad-slot.porvata-outstream {
+	&::before {
+		content: attr(data-label);
+		display: block;
+		font-size: 0.625rem;
```
dianafa: we can make it `.625rem;` as well as `.75rem;` below.

Why do we want to use `rem` here instead of px like in all other places?

dianafa (approved):
One comment, rest LGTM

PR #157, opened 2017-09-18
XW-3939 | Single file that can work for server and fastboot-server fo…

hakubo:
@Wikia/x-wing @rogatty 

I need to rename server folder
otherwise ember serve fails

but to do that I need to update path in chef-repo
https://github.com/Wikia/chef-repo/blob/master/cookbooks/mobile-wiki/templates/default/sv-mobile-wiki-run.erb#L33

### DIFF
--- a/fastboot-server.js
+++ b/fastboot-server.js
@@ -1,0 +1,12 @@
+/*
+Temporary file needed for XW-3939
+to be able to run ember serve
+we can't have serve/server files/folders in root directory
+
+We do have hardcoded path to server/app/server in chef repo though
+so to enable us to be able to change name of the 'server' folder
+I need to have a single file that would work for both cases
+to be able to change that in chef repo
+ */
+
+require('./server/app/server');


rogatty (approved):

rogatty:
@kvas-damian do you remember the reason why we used `server/app/server.js` and not `server.js` as we wanted initially? There was _some_ reason...

hakubo:
I think server.js would not work as well
as 
ember serve
would like to import it
and it would die

rogatty:
I don't think it was that. I'd try `fastboot-server.js` before `server/app/server.js` ;) I believe there was some other reason but can't remember it.

rogatty:
Anyway @hakubo, I wouldn't worry too much about it. The current change shouldn't break anything, and in the future you'll test what happens when you change the path in chef.

kvas-damian:
@rogatty AFAIR it was related to backwards compatibility. We were introducing a lot of changes in FastBoot project and we wanted to avoid this one (code on old branches won't work now). Also we were running fastboot and non-fastboot version in parallel, so we needed server runner to work the same on both platforms.

rogatty:
>we were running fastboot and non-fastboot version in parallel

Oh, I think it was that, thanks!

PR #161, opened 2017-09-22
XW-3998 | Moving mobile-wiki to k8s

xkxd:
## Links

https://wikia-inc.atlassian.net/browse/XW-2978 + subtickets - main ticket
https://wikia-inc.atlassian.net/browse/XW-3998 - dockerfile related ticket

## Description
- dockerfiles for dev and prod
- k8s yaml descriptor templates
- bash scripts for jenkins
- jenkinsfile
- automated by Jenkins Jobs Builder

🎉 

## Reviewers
@Wikia/x-wing


config/environment.js
```diff
@@ -100,7 +100,7 @@ module.exports = function (environment) {
 			// when it is created
 		},
 		fastboot: {
-			hostWhitelist: [/.*\.wikia-dev\.(pl|us)/, /.*\.wikia-staging.com/, /.*\.wikia.com/],
+			hostWhitelist: [/.*\.wikia-dev\.(pl|us)/, /.*\.wikia-staging.com/, /.*\.wikia.com/,  /^localhost:\d+$/],
```
xkxd: it's related to an error occurring while trying to run docker image locally, not sure if we want it now (or ever)

Dockerfile
```diff
@@ -14,8 +14,54 @@ RUN useradd -u 663 release
...
+# store prod dependencies separately for docker's caching reasons
+RUN mkdir prod_dependencies
+RUN cp -R node_modules prod_dependencies
+RUN cp -R bower_components prod_dependencies
+
+# install all dependencies
+RUN npm run setup
+
+# copy app
+COPY . .
```
bkoval: Could we add a comment explaining the dots?

bkoval (approved):

Dockerfile
```diff
@@ -14,8 +14,60 @@ RUN useradd -u 663 release
...
-COPY . /build
-WORKDIR /build
+# create folder to cache dependencies
+WORKDIR /app
+COPY package.json .
+COPY bower.json .
+
+# install npm globals
+RUN npm install -g bower
+RUN npm install -g ember-cli
```
rogatty: These two shouldn't be required as we don't run them directly but through npm scripts which use the local version anyway.

Dockerfile
```diff
@@ -14,8 +14,60 @@ RUN useradd -u 663 release
...
+RUN npm run clean-deep
```
slayful: Is this necessary?

xkxd: no, it's not

Dockerfile.build
```diff
@@ -0,0 +1,56 @@
...
+        git
+
+# create folder to cache dependencies
+WORKDIR /app
+COPY package.json .
+COPY bower.json .
+
+# install npm globals
+RUN npm install -g bower
+RUN npm install -g ember-cli
```
rogatty: How about using `npm run setup-for-local`? We should lock the versions and it's done there.

tasks/docker-build.sh
```diff
@@ -0,0 +1,39 @@
...
+docker cp temporary-container:/app/config  ./docker_temp_container
+docker cp temporary-container:/app/lib  ./docker_temp_container
+docker cp temporary-container:/app/package.json  ./docker_temp_container
+docker cp temporary-container:/app/bower.json  ./docker_temp_container
+
+docker cp temporary-container:/app/prod_dependencies/node_modules ./docker_temp_container
+docker cp temporary-container:/app/prod_dependencies/bower_components ./docker_temp_container
+
+# temporary container is not needed anymore so we're removing it
+docker rm -f temporary-containergit
```
slayful: Is that correct?

Dockerfile.build
```diff
@@ -0,0 +1,56 @@
...
+# - second argument is the docker daemon (newly created container)
+# - "." means "current working directory", and since we're not selecting any specific files
+# it's basically "copy everything from mobile-wiki project to WORKDIR in docker"
+COPY . .
+
+# build app
+RUN npm run build-prod
+
+RUN apk del .gyp python make g++ git
+RUN npm uninstall -g bower
```
rogatty: According to https://www.ctl.io/developers/blog/post/optimizing-docker-images/#image-size it doesn't make sense as the final image size won't shrink.

Dockerfile.build
xkxd: we don't have it in `mobile-wiki` but thanks, worth considering :)

app/services/logger.js
```diff
@@ -74,17 +74,16 @@ export default Service.extend({
 
 	setupBunyan() {
 		const bunyan = FastBoot.require('bunyan');
-		const BunyanSyslog = FastBoot.require('bunyan-syslog');
+		const process = FastBoot.require('process');
+
 		const instance = bunyan.createLogger({
 			appname: 'mobile-wiki',
 			name: 'mobile-wiki',
+			loggerName: 'services/logger.je',
```
slayful: perhaps `logger.js`?

k8s/k8s-descriptor-template-preview.yaml
```diff
@@ -0,0 +1,76 @@
...
+        env:
+          - name: WIKIA_ENVIRONMENT
+            value: prod
+          - name: WIKIA_DATACENTER
+            value: sjc
+          - name: MEDIAWIKI_DOMAIN
+            value: prod.icache.service.sjc.consul
+          - name: GA_USERID_SALT
+            valueFrom:
+              secretKeyRef:
```
rogatty: We should use https://github.com/Wikia/vault-init-container instead.

PR #192, opened 2017-10-10
Run tests in headless chrome

hakubo:
PhantomJS is not maintained anymore and has its drawbacks
Let's move to headless chrome

https://wikia-inc.atlassian.net/browse/PLATFORM-3223


Dockerfile
```diff
@@ -1,9 +1,5 @@
 FROM node:6.10
 
-# phantomjs workaround
```
xkxd, line 3: note: we have to delete it in https://github.com/Wikia/mobile-wiki/pull/161/files#diff-3254677a7917c6c01f55212f86c57fbfR10 as well

xkxd (approved):

rogatty (approved):
```
ok 17 Chrome 62.0 - Acceptance | search: visiting /search
    ---
        Log: |
            { type: 'warn',
              text: '\'You created a second Pretender instance while there was already one running. Running two Pretender servers at once will lead to unexpected results and will be removed entirely in a future major version.Please call .shutdown() on your instances when you no longer need them to respond.\'\n' }
```
is alarming but other than that, LGTM

latata (approved):

PR #198, opened 2017-10-17
Disable Linter on Jenkins

hakubo:
it runs with ember test anyway

## Description

Ember recently moved natively to eslint so it runs with ember test command by default.
That means we currently run linting twice on each PR, this PR removes standalone linter on Jenkins.


rogatty:
How about removing the npm script too?

hakubo:
I did not remove it now so we can run it manually.

rogatty:
I noticed it doesn't lint `config/*`, `fastboot-server/*`, `lib/*` and a few others. Is it possible to add them to the list?

PR #201, opened 2017-10-18
don't lint files twice, most of files are linted by ember test

hakubo:
## Links

We cannot do https://github.com/Wikia/mobile-wiki/pull/198
as some files would not be linted

## Description

This PR disables npm run linter from linting all files
majority of them are being linted by ember test


rogatty (approved):

rogatty:
LGTM but it might be confusing when people see that linter passes, tests fail and the fix is to change the indentation 😉

hakubo:
yup, but there is no way to add additional files to ember-eslint now

PR #202, opened 2017-10-18
XW-4110 | Cookie for all subdomains, cleanups, cookie setting logic

bkoval:
## Links

* http://wikia-inc.atlassian.net/browse/XW-4110

## Reviewers

@Wikia/x-wing

### DIFF
--- a/app/components/fandom-app-smart-banner.js
+++ b/app/components/fandom-app-smart-banner.js
@@ -15,12 +15,13 @@ export default Component.extend({
 
 	options: {
 		// Duration to hide the banner after close button is clicked (0 = always show banner)
-		daysHiddenAfterClose: 15,
+		daysHiddenAfterClose: 30,
 
 		// Duration to hide the banner after it is clicked (0 = always show banner)
-		daysHiddenAfterView: 30,
+		daysHiddenAfterView: 90,
 	},
-	dayInSeconds: 86400000,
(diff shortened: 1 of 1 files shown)


app/components/fandom-app-smart-banner.js
```diff
@@ -37,6 +38,18 @@ export default Component.extend({
...
+	/**
+	 * Obtains domain we should cookies on. Since Fandom App SmartBanner is global for all the wikis,
+	 * we should set it to all subdomains of wikia.com, wikia-dev.us or wikia.dev-pl
+	 *
+	 * @returns {string}
+	 */
+	cookieDomain: computed(() => {
+		const domainSegmentsReversed = window.location.hostname.split('.').reverse();
+
+		return `.${domainSegmentsReversed[1]}.${domainSegmentsReversed[0]}`;
```
rogatty: You can get `cookieDomain` from config which is the same thing. See https://github.com/Wikia/mobile-wiki/blob/de66e98fd25b61ab9586abdb4765cb83a9d77208/app/components/global-footer/global-footer-bottom-bar.js#L14

hakubo: +1

hakubo (changes requested):

latata (approved):

hakubo (approved):

PR #215, opened 2017-11-02
XW-4146 | Log connection types to inform future autoplay decisions

JamesHawking:
## Links

* https://wikia-inc.atlassian.net/browse/XW-4146

## Description

Log the connection type via Google Analytics for mobile-wiki.

## Reviewers

@Wikia/x-wing


app/mixins/connection-type.js
```diff
@@ -0,0 +1,19 @@
...
+const {Mixin, inject} = Ember;
+
+export default Mixin.create({
+	fastboot: inject.service(),
+	connectionType: 'unresolved',
+
+	init() {
+		if (!this.get('fastboot.isFastBoot')) {
+			const connection = navigator.connection || navigator.mozConnection || navigator.webkitConnection;
+			const effectiveConnectionType = connection.effectiveType || connection.type;
```
latata: I'd only track only effectiveType as simple type doesn't give us any value.

xkxd: i'd put this line into `if (connection)` check
\+ variable will be redundant then

hakubo: is effectiveType the same as type?
if not, I don't think we should log it as one value to not confuse us

app/controllers/application.js
```diff
@@ -62,6 +65,12 @@ export default Controller.extend(
 			});
 
 			this._super();
+
+			track({
+				action: trackActions.open,
+				category: 'connection-type',
+				label: `mobile-${this.get('connectionType')}`
```
xkxd: consider sth like `Mobile: ${this.get('connectionType')}`

xkxd (approved):

app/controllers/application.js
```diff
@@ -62,6 +65,12 @@ export default Controller.extend(
 			});
 
 			this._super();
+
+			track({
```
hakubo: don't log if unresloved connectionType

app/controllers/application.js
latata: `mobile-` prefix is not necessary. By default we add `mercury-`  prefix to category in mobile-wiki

hakubo: I don't think we need to have mobile- in label
we used to add mobile-wiki to every category

app/mixins/connection-type.js
latata: if `connection` is null exception will be thrown

app/controllers/application.js
```diff
@@ -62,6 +65,12 @@ export default Controller.extend(
...
+				action: trackActions.open,
```
hakubo: more like view action?

app/mixins/connection-type.js
JamesHawking: Check out @xkxd comment.

app/controllers/application.js
```diff
@@ -62,6 +65,12 @@ export default Controller.extend(
...
+				category: 'connection-type',
+				label: `mobile-${this.get('connectionType')}`
+			});
```
hakubo: this event is nonInteractive
https://github.com/Wikia/mobile-wiki/blob/dev/app/utils/track.js#L127

app/controllers/application.js
JamesHawking: 👍

app/mixins/connection-type.js
xkxd: @hakubo , maybe we should log both, since "effective" means "4g / 3g" and "type" means "wifi / cellular" ?

app/mixins/connection-type.js
```diff
@@ -0,0 +1,19 @@
...
+
+			if (connection && effectiveConnectionType) {
+				this.set('connectionType', effectiveConnectionType);
```
latata: I'd make this as computed property.

latata (changes requested):

app/mixins/connection-type.js
hakubo: yeah, logging both make sense to me

app/mixins/connection-type.js
```diff
@@ -0,0 +1,19 @@
+import Ember from 'ember';
+
+const {Mixin, inject} = Ember;
+
+export default Mixin.create({
+	fastboot: inject.service(),
+	connectionType: 'unresolved',
+
+	init() {
+		if (!this.get('fastboot.isFastBoot')) {
```
bkoval: What do you think about using a computed property instead of "init"?

E.g. connectionType would be a computed property that recalculates on 'fastBoot.isFastboot'.

What it gives us? No race conditions: if you fetch a computed property using a getter, you are sure the logic below was or will be executed, no matter where you use the mixin :)

app/controllers/application.js
bkoval: Shouldn't we make some sort of check if connectionType exists before sending a tracking request?

JamesHawking: Can you be more specific with this one? :)

xkxd: I wanted to get rid of "-", but others convinced me to remove "mobile" completely, that's why you can skip this comment :D

hakubo (approved):

app/mixins/connection-type.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+const {Mixin, inject, computed} = Ember;
+
+export default Mixin.create({
+	fastboot: inject.service(),
+	connection: computed('fastboot.isFastBoot', function () {
+		if (!this.get('fastboot.isFastBoot')) {
+			return navigator.connection || navigator.mozConnection || navigator.webkitConnection;
+		} else {
```
latata: I'd remove this `else`

app/mixins/connection-type.js
```diff
@@ -0,0 +1,17 @@
...
+			return 'unresolved';
+		}
+	}),
+	effectiveConnectionType: computed('connection', function () {
+		return this.get('connection.effectiveType') || this.get('connection.type');
```
latata: I'd remove `connection.type`. I believe it can cause confusion.

latata (changes requested):

app/mixins/connection-type.js
JamesHawking: It can be simply filtered in GA, and still can provide some data.

app/mixins/connection-type.js
JamesHawking: Does computed props defaults to false if there is no return?

xkxd: i'd assume that it'll be equal to `undefined` if there is no return

latata (approved):

app/mixins/connection-type.js
```diff
@@ -0,0 +1,15 @@
...
+const {Mixin, inject, computed} = Ember;
+
+export default Mixin.create({
+	fastboot: inject.service(),
+	connection: computed('fastboot.isFastBoot', function () {
+		if (!this.get('fastboot.isFastBoot')) {
+			return navigator.connection || navigator.mozConnection || navigator.webkitConnection;
+		}
+	}),
+	effectiveConnectionType: computed('connection', function () {
```
bkoval, line 12: I don't think we need both computed properties. I think all the logic could be encapsulated in one computed property (effectiveConnectionType) since we do not need connection itself :)

JamesHawking: Don't you think is more readable this way?

latata: I agree with @bkoval . There is not need to add extra computed property.

latata (approved):

PR #223, opened 2017-11-03
XW-4191: Unmute property

JamesHawking:
## Links

https://wikia-inc.atlassian.net/browse/XW-4191

## Description

As a user who clicks/taps a video to play in recirculation, the video should be unmuted

## Reviewers

@Wikia/x-wing


latata (approved):
We will need to set a proper jwplayer-fandom version to make it work.

rybmat (approved):

PR #228, opened 2017-11-07
XW-3972 | JW Player parser tag

bkoval:
@Wikia/x-wing


app/components/jwplayer-embed.js
```diff
@@ -0,0 +1,44 @@
+import Ember from 'ember';
```
latata: I'd name the file `jwplayer-tag` to be consistent with App.

app/components/jwplayer-embed.js
```diff
@@ -0,0 +1,44 @@
...
+			window.wikiaJWPlayer(
+				this.get('jwPlayerRootId'),
+				this.getPlayerSetup(videoData)
+			);
+		});
+	},
+
+	getPlayerSetup(jwVideoData) {
+		return {
+			autoplay: {
```
latata: API has changed. Please merge dev and update to `autoplay: false`

app/modules/jwplayer-assets.js
```diff
@@ -0,0 +1,47 @@
+import Ember from 'ember';
+const { RSVP } =  Ember;
+
+export const assetUrls = {
```
latata: Do we need to export it?

app/modules/jwplayer-assets.js
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	load() {
+		this.loadStyles();
+
+		return this.loadScripts();
+	}
+}
+
+const jwPlayerAssets = new JWPlayerAssets();
```
latata: Maybe we should create class with only static methods?

app/components/jwplayer-embed.js
latata: After merging dev you will also need to update imports, am I right?

app/components/jwplayer-embed.js
```diff
@@ -0,0 +1,44 @@
...
+
+	jwPlayerRootId: computed('media-id', function () {
+		return `jwPlayerTag${this.get('media-id')}`;
+	}),
+
+	/**
+	 * @returns {void}
+	 */
+	didInsertElement() {
+		RSVP.Promise.all([
```
rogatty: Remember about calling `this._super` https://stackoverflow.com/a/41011291

app/components/jwplayer-embed.js
```diff
@@ -0,0 +1,44 @@
...
+			jwPlayerAssets.load(),
+			this.getVideoData()
+		]).then(([, videoData]) => {
```
rogatty: If you use `hash` instead of `all` you won't need this empty space. See https://github.com/tildeio/rsvp.js/#hash-of-promises

app/components/jwplayer-embed.js
```diff
@@ -0,0 +1,44 @@
...
+			window.wikiaJWPlayer(
+				this.get('jwPlayerRootId'),
```
rogatty: If a user navigates away from the page before this is called, an error will be thrown. Check for `this.isDestroyed()`

app/modules/jwplayer-assets.js
```diff
@@ -0,0 +1,48 @@
+import Ember from 'ember';
+
+const {RSVP} = Ember;
+
+export const assetUrls = {
+	styles: '/mobile-wiki/assets/jwplayer/index.css',
+	script: '/mobile-wiki/assets/jwplayer/wikiajwplayer.js'
```
rogatty: Remember to test on a sandbox if these are fingerprinted correctly. There were issues in the past.

app/modules/jwplayer-assets.js
```diff
@@ -0,0 +1,48 @@
...
+	load() {
+		this.loadStyles();
+
+		return this.loadScripts();
+	}
+}
+
+const jwPlayerAssets = new JWPlayerAssets();
+
+export default jwPlayerAssets;
```
xkxd: variable is redundant

app/modules/video-players/jwplayer.js
```diff
@@ -87,16 +83,9 @@ export default class JWPlayer extends BasePlayer {
...
-	}
-
-	loadStyles(cssFile) {
-		$(`<link rel="stylesheet" href="${cssFile}" crossorigin="anonymous">`).appendTo('head');
-	}
-
-	loadScripts(jsFile, callback) {
-		window.M.loadScript(jsFile, true, callback, 'anonymous');
+		JWPlayerAssets.load().then((data) => {
+			this.playerDidLoad.bind(this)(data);
```
xkxd: you're using arrow func, no need to bind

xkxd (changes requested):

app/components/jwplayer-tag.js
```diff
@@ -0,0 +1,51 @@
+import Ember from 'ember';
```
latata: change to new ember imports

app/components/jwplayer-tag.js
```diff
@@ -0,0 +1,51 @@
...
+import jwPlayerAssets from '../modules/jwplayer-assets';
+import {track} from '../utils/track';
+
+const {RSVP, $, computed, Component} = Ember;
```
latata: change to new ember imports

app/modules/jwplayer-assets.js
bkoval: 👍

app/modules/jwplayer-assets.js
bkoval: As we talked offline, we'll leave it as is :)

latata (approved):

xkxd (approved):

PR #240, opened 2017-11-10
XW-4045 | Upgrade to node 8.9.1

hakubo:


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "ember build",
     "build-prod": "ember build -prod",
     "clean": "rm -rf node_modules/ bower_components/ dist/ tmp/",
-    "clean-deep": "npm run clean && npm cache clean && bower cache clean",
```
rybmat, line 14: why `npm cache clean` is not needed anymore?

fastboot-server/app/server.js
```diff
@@ -1,22 +1,3 @@
...
- */
-if (process.env.HEAPDUMP_ENABLED === 'true') {
-	require('heapdump');
-}
-
-/**
- * NewRelic is only enabled on one fastboot-server and that logic is managed by chef,
- * which passes it to our config
- */
-if (process.env.NEW_RELIC_ENABLED === 'true') {
```
rybmat, line 16: 👍

app/styles/app.scss
```diff
@@ -1,14 +1,14 @@
...
 // load order of files below is intentional
-@import "../../bower_components/wikia-style-guide/src/scss/normalize",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/global",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/buttons",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/dropdown",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/forms",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/imagery",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/pagination",
-"../../bower_components/wikia-style-guide/src/scss/lib/components/type";
-@import '../../bower_components/design-system/dist/scss/index';
```
rybmat, line 11: what has changed that we don't need this `../../`?

package.json
hakubo: npm 5  guarantees cache to be in good state

rybmat (approved):

fastboot-server/app/server.js
hakubo: 👍

app/styles/app.scss
hakubo: I pass what files to include in sassOptions in ember-cli file

Dockerfile.dev
```diff
@@ -28,7 +29,7 @@ RUN echo '{ "allow_root": true }' > /root/.bowerrc && \
     apk del .gyp python make g++ git && \
     bower cache clean && \
     npm uninstall -g bower && \
-    npm cache clean
+    npm cache clean --force
```
xkxd, line 32: why?

Dockerfile.prod
```diff
@@ -17,18 +17,14 @@ RUN echo '{ "allow_root": true }' > /root/.bowerrc && \
 # install missing stuff
     apk add --no-cache --virtual .gyp python make g++ git && \
 
-# install npm globals
-    npm install -g bower && \
```
xkxd, line 21: previously `npm run setup` tend throw in case of missing bower, is it taking the local one now?

xkxd (approved):

Dockerfile.dev
hakubo: https://docs.npmjs.com/cli/cache#details

npm stores cache data in an opaque directory within the configured cache, named _cacache. This directory is a cacache-based content-addressable cache that stores all http request data as well as other package-related data. This directory is primarily accessed through pacote, the library responsible for all package fetching as of npm@5.

All data that passes through the cache is fully verified for integrity on both insertion and extraction. Cache corruption will either trigger an error, or signal to pacote that the data must be refetched, which it will do automatically. For this reason, it should never be necessary to clear the cache for any reason other than reclaiming disk space, thus why clean now requires --force to run.

There is currently no method exposed through npm to inspect or directly manage the contents of this cache. In order to access it, cacache must be used directly.

npm will not remove data by itself: the cache will grow as new packages are installed.

Dockerfile.prod
hakubo: https://github.com/Wikia/mobile-wiki/pull/240/files#diff-0e9722a692af2b244674b56146a1ecd3R26

hakubo: and when scripts are run via "npm run" they have .bin added to path

PR #241, opened 2017-11-13
Log app_version

hakubo:
This takes a form of
0.0.0-commitHash

0.0.0 is coming from package.json version

This way will be able to see in logs what version of the application was running the code

app_name is already logged so it is not needed in @ fields


latata (approved):

rybmat (approved):
@hakubo remember to create a ticket for unifying loggers

PR #244, opened 2017-11-14
XW-4268 | send logs to event-logger service in Ember.onerror

rybmat:
## Links

* https://wikia-inc.atlassian.net/browse/XW-4268

## Description
To gather more data about bug from ticket, we'll send ember errors to event-logger service

## Reviewers

@Wikia/x-wing

### DIFF
--- a/app/instance-initializers/error-logger.js
+++ b/app/instance-initializers/error-logger.js
@@ -1,4 +1,6 @@
 import config from '../config/environment';
+import fetch from 'fetch';
+import Ember from 'ember';
 
 export function initialize(/* appInstance */) {
 	if (typeof FastBoot !== 'undefined') {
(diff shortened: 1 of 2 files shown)


app/instance-initializers/error-logger.js
```diff
@@ -0,0 +1,26 @@
...
+
+export function initialize(/* appInstance */) {
+	if (typeof FastBoot !== 'undefined') {
+		return;
+	}
+
+	Ember.onerror = function (error) {
+		const url = `https://${config.services.domain}/${config.services.eventLogger.baseAPIPath}/error`;
+
+		Ember.$.ajax(url, {
```
hakubo: use fetch for consistency

app/instance-initializers/error-logger.js
```diff
@@ -0,0 +1,26 @@
...
+				client: "mobile-wiki"
+			}
+		});
+	};
+}
+
+export default {
+	after: 'config',
+	name: 'error-logger',
+	initialize
```
xkxd, line 30: afair airbnb style guide says that shorthand props should be at the top of object

xkxd (approved):

PR #251, opened 2017-11-15
Remove application-wrapper

hakubo:
This wrapper adds nothing but confusion

there is another one - article-wrapper - but that will be harder to remove

as per: https://guides.emberjs.com/v2.16.0/controllers/#toc_common-questions


xkxd (approved):
I assumed you didn't change anything besides coping code from file to file.

PR #260, opened 2017-11-20
XW-2830 | Make sure that Local Nav Links Expressed in HTML

latata:
https://wikia-inc.atlassian.net/browse/XW-2830
@Wikia/x-wing

Want to check if it cause memory leak after ember/fastboot update. Hopefully not.
#116

### DIFF
--- a/app/components/fastboot-only/local-navigation-for-seo.js
+++ b/app/components/fastboot-only/local-navigation-for-seo.js
@@ -10,7 +10,7 @@ export default Component.extend({
 	currentLocalLinks: or('currentLocalNav.children', 'localLinks'),
 	flattenLinksArray(linksList) {
 		let flatArray = [];
-		linksList.forEach(item => {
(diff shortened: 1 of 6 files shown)


app/components/fastboot-only/local-navigation-for-seo.js
```diff
@@ -0,0 +1,26 @@
...
+			if (item.href !== '#') {
+				flatArray.push({name: item.text, href: item.href});
+			}
+			if (item.children) {
+				flatArray = flatArray.concat(this.flattenLinksArray(item.children));
+			}
+		});
+		return flatArray;
+	},
+	model: computed('flattenLinksArray', function () {
```
hakubo: computed('currentLocalLinks')

hakubo (changes requested):

hakubo (approved):

PR #281, opened 2017-11-27
XW-4302 | Use template lint fully

hakubo:
Changes:
- use dot reporter by default
- enable recommended rules for template linter
- disabled some that would be too much work now to fix


JamesHawking (approved):
LGTM

rybmat (approved):

app/templates/components/lightbox-image.hbs
```diff
@@ -1,6 +1,7 @@
-{{#unless loadingError}}
-	<img class="current" src="{{imageSrc}}" style={{style}}>
-{{else}}
+{{#if loadingError}}
 	{{i18n 'app.media-lightbox-error'}}
+{{/if}}
+{{#unless loadingError}}
```
rogatty: How about `{{#if}}{{else}}{{/if}}`?

rogatty (approved):
Good stuff :+1:

app/templates/components/article-comment.hbs
```diff
@@ -1,5 +1,5 @@
 <div class='avatar'>
```
latata, line 1: Can it check if only double quotes are used for attribute values?

hakubo: it will be able in next version

PR #292, opened 2017-12-04
XW-4268 | Null-check for googletag.pubads

latata:
https://wikia-inc.atlassian.net/browse/XW-4268
@Wikia/x-wing


app/components/ad-slot.js
```diff
@@ -94,7 +95,13 @@ export default Component.extend(
 			const name = this.get('name');
 
 			this.get('logger').info('Will destroy ad:', name);
-			this.get('ads.module').removeSlot(name);
+			// XW-4268 - temporary fix that prevents interrupting glimmer transaction,
+			// when exception of pubads being undefined is thrown
+			if (window.googletag && window.googletag.pubads) {
+				this.get('ads.module').removeSlot(name);
+			} else {
+				logEvent('window.googletag.pubads is undefined');
```
rybmat: this functions takes two parameters, event name and event description. Can we provide both of them?

rybmat (approved):

rybmat (approved):

PR #319, opened 2017-12-11
XW-4383 | load assets from the same domain

hakubo:
This does mean we'll have worse cache hit ratio and we will load more data due to sending cookies with the request

but not having to resolve another domain, creating new TLS connection can save as even ~600ms. 

Once we'll switch to HTTPS on main domains - this would be even faster, as it will be able to reuse the same single TCP connection

Comparison:
http://www.webpagetest.org/video/compare.php?tests=171219_NZ_f62090e68f7b5d24407ae0a2d314e430%2C171219_34_789316a6633b1480e5aeedc4f0d7639b&thumbSize=200&ival=100&end=visual


latata (approved):

rybmat (approved):

rogatty:
Which cookies are added when we stop using `nocookie.net`? I checked the VCL logic and found only this:
https://github.com/Wikia/wikia-vcl/blob/9712150b40f52e03aca24ef376abfe86b856356e/common/common-service-beacon.vcl#L47-L70

If the cookies are a problem, I'm sure there is a way to use the same cleaning process as on `nocookie.net`

PR #332, opened 2017-12-20
XW-4393 | Collapsible sections

latata:
https://wikia-inc.atlassian.net/browse/XW-4393
@Wikia/x-wing

Wait with releasing this at least 24 hours since https://github.com/Wikia/app/pull/14442 is released


rybmat (approved):

app/styles/module/article/_section-header.scss
```diff
@@ -1,21 +1,32 @@
...
 	}
 
-	.chevron {
-		transform: rotate(180deg);
+	.contribution-container {
+		visibility: hidden;
 	}
 }
+
+.mobile-hidden {
```
hakubo, line 24: there is global hidden class that you can use

app/utils/render-component.js
```diff
@@ -46,14 +46,28 @@ export function getRenderComponentFor(parent) {
...
+}
+
+export function queryPlaceholders(element, filterHidden) {
 	const components = [];
+	let componentElements = element.querySelectorAll('[data-component]');
 
-	$element.find('[data-component]').each(function () {
-		const name = this.getAttribute('data-component'),
-			attrs = componentAttributes(this);
+	componentElements.forEach((componentElement) => {
```
hakubo, line 53: What do you think about using https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent
this should return null if element is not visible
something like this:
[...componentElements].filter((componentElement) => componentElement.parentOffset)

app/styles/module/article/_section-header.scss
latata: `.wds-is-hidden` is defined with `!important`, I wouldn't like to override `!important` with more important `!important` ;)

hakubo (approved):

app/modules/ads.js
```diff
@@ -332,28 +332,25 @@ class Ads {
...
 		if (this.getTargetingValue('pageType') === 'home') {
-			return hasTrendingArticles;
+			return $('.trending-articles').length > 0;
 		}
 
-		return hasArticleFooter && !showInContent || articleBodyHeight > this.adsData.minPageLength;
+		const numberOfSections = $('.article-content > h2').length,
+			hasArticleFooter = $('.article-footer').length > 0;
+
+		return hasArticleFooter && (!isInContentApplicable || numberOfSections > this.adsData.minPageLength);
```
kvas-damian:
```
0&&1||1
//1
0&&(1||1)
//0
```
Are you aware that you're changing the logic by adding parenthesis?

PR #333, opened 2017-12-20
XW-4290 | Logging app version

latata:
https://wikia-inc.atlassian.net/browse/XW-4290
@Wikia/x-wing

### DIFF
--- a/app/modules/event-logger.js
+++ b/app/modules/event-logger.js
@@ -14,7 +14,7 @@ export default function logEvent(name, description) {
 		body: JSON.stringify({
 			name,
 			description: JSON.stringify(description),
-			client: 'mobile-wiki'
+			client: `mobile-wiki:${config.APP.version}`
 		})
 	});
 }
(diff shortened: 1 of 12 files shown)


app/modules/event-logger.js
```diff
@@ -14,7 +14,7 @@ export default function logEvent(name, description) {
 		body: JSON.stringify({
 			name,
 			description: JSON.stringify(description),
-			client: 'mobile-wiki'
+			client: `mobile-wiki:${config.APP.version}`
```
rybmat: How about separate (optional) field called `client_version` in the object sent to event-logger? It is 10 minutes of work on the service side and would be more consistent with the approach we have in other applications

app/modules/video-players/jwplayer.js
```diff
@@ -74,7 +74,7 @@ export default class JWPlayer extends BasePlayer {
 					playlist: this.params.playlist
 				},
 				logger: {
-					clientName: 'mobile-wiki'
+					clientName: `mobile-wiki:${config.APP.version}`
```
rybmat: like above

rybmat (changes requested):

rybmat (approved):

PR #337, opened 2017-12-28
XW-4375 | enable order-in-controllers in eslint-plugin-ember

latata:
https://wikia-inc.atlassian.net/browse/XW-4375
@Wikia/x-wing


rybmat (approved):

hakubo (changes requested):
package-lock is changed, why?

hakubo (approved):

PR #359, opened 2018-01-19
PLATFORM-3314: content security policy header for https requests

themech:
## Links

* https://wikia-inc.atlassian.net/browse/PLATFORM-3314
* https://github.com/Wikia/csp-logger

## Description

We want to know what's broken for HTTPS requests. Here we're adding csp response header so all the ajax and assets requests not compatible with our policy will be reported. For now it is only reporting, enforcing and `Upgrade-Insecure-Requests` will come later.

## Reviewers

FYI: @Wikia/x-wing


Grunny (approved):

hakubo (changes requested):
There is a standard library in Ember to do exactly this
https://github.com/rwjblue/ember-cli-content-security-policy

themech:
@hakubo I'm not sure if we want to use the library. The change is pretty small and also we want to emit this header only when request contains the `fastly-ssl` header. By briefly going through the standard lib doc it looks like it is configured for the whole app, not per request.

hakubo (approved):
ok!

PR #391, opened 2018-02-02
XW-4592 | Removing $.extend occurrences

bkoval:
@Wikia/x-wing

### DIFF
--- a/app/utils/extend.js
+++ b/app/utils/extend.js
@@ -1,5 +1,4 @@
-function deepExtend(out) {
-	out = out || {};
+function deepExtend(out = {}) {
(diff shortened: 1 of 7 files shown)


rybmat (approved):

app/utils/extend.js
```diff
@@ -1,8 +1,30 @@
-// We can't use Object.assign because it's not doing deep merge
+function deepExtend(out) {
```
latata: We can use ES6 default argument value `out = {}`

latata (approved):

hakubo (approved):

PR #394, opened 2018-02-02
XW-4594 | Removing $ from BasePlayer

bkoval:
@Wikia/x-wing

### DIFF
--- a/app/modules/video-players/base.js
+++ b/app/modules/video-players/base.js
@@ -54,10 +54,10 @@ export default class BasePlayer {
 	 * @returns {void}
 	 */
 	onResize(containerSelector = this.containerSelector) {
-		const $container = $(containerSelector),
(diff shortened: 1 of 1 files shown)


app/modules/video-players/base.js
```diff
@@ -80,7 +80,7 @@ export default class BasePlayer {
 			};
 		}
 
-		$container.css(sanitizedSize);
+		Object.assign(container.style, sanitizedSize);
```
latata, line 83: did you test it? It might not work as `sanitizedSize` contains integer values, not string with `px` suffix. I'm just guessing...

bkoval: Funny, I did test that and it wasn't spitting any errors. I will add the 'px' though :)

latata (approved):

hakubo (approved):

PR #401, opened 2018-02-05
XW-4155 | Remove jquery from mobile-wiki

hakubo:


app/modules/jwplayer-assets.js
```diff
@@ -16,8 +16,12 @@ class JWPlayerAssets {
 
 	loadStyles() {
 		if (!this.wasStyleLoadInitialized) {
-			$(`<link rel="stylesheet" href="${assetUrls.styles}" crossorigin="anonymous">`)
-				.appendTo('head');
+			const styles = document.createElement('link');
+
+			styles.rel = 'stylesheet';
+			styles.href = assetUrls.styles;
+			document.head.appendChild(styles);
```
latata, line 23: Missing `crossorigin`?

app/components/recirculation-prefooter.js
```diff
@@ -40,13 +40,16 @@ export default Component.extend(
...
+			const viewportTolerance = 100;
+
+			this.setProperties({
+				viewportTolerance: {
+					top: viewportTolerance,
+					bottom: viewportTolerance,
+					left: 0,
+					right: 0
+				},
+				intersectionThreshold: 0
```
hakubo, line 52: shouldn't that be default?

latata: for some reason default is `1`

PR #410, opened 2018-02-07
XW-4602 | Removed site-head-fandom-bar

kolek7:
## Links

https://wikia-inc.atlassian.net/browse/XW-4602

## Reviewers

@Wikia/x-wing


app/components/site-head.js
```diff
@@ -23,7 +23,7 @@ export default Component.extend(
 		offset: 0,
 
 		defaultWikiaHomePage: 'http://fandom.wikia.com',
-
+		partnerSlot: alias('globalNavigation.partner_slot'),
```
latata: use `readOnly` instead of `alias`. `alias` gives two-way binding which we don't need here.

app/components/category-members-grouped.js
```diff
@@ -30,7 +30,7 @@ export default Component.extend(
 
 				this.get('loadPage')(page)
 					.then(() => {
-						const navHeight = $('.site-head').outerHeight() + $('.site-head-fandom-bar').outerHeight(),
+						const navHeight = $('.site-head').outerHeight(),
```
rybmat: @kolek7 I think we should change it to check outerHeight of `.site-head-wrapper` to handle german wikis well

rybmat: BTW, we should also replace jquery here with plain javascript code, as we are getting rid of jquery in mobile-wiki at all

rybmat (dismissed):

app/styles/component/_site-head.scss
```diff
@@ -149,3 +157,32 @@ body:not(.no-scroll) {
...
+	top: 55px;
+	width: 100%;
+	z-index: 695;
+
+	.featured-video-on-scroll-active & {
+		position: static;
+	}
+}
+
+.site-head-fandom-bar__partner-logo {
```
rybmat: we can skip `fandom-bar` part from this class name

app/styles/component/_site-head.scss
```diff
@@ -149,3 +157,32 @@ body:not(.no-scroll) {
...
+	background: white;
+	border-bottom: 1px solid $vapor;
+	height: $partner-slot-height;
+	position: fixed;
+	top: 55px;
+	width: 100%;
+	z-index: 695;
+
+	.featured-video-on-scroll-active & {
+		position: static;
```
rybmat: we should apply `position: static` also if `bfaa-template` class is applied, see L148 and L156

app/components/category-members-grouped.js
```diff
@@ -30,7 +30,7 @@ export default Component.extend(
 
 				this.get('loadPage')(page)
 					.then(() => {
-						const navHeight = $('.site-head').outerHeight() + $('.site-head-fandom-bar').outerHeight(),
+						const navHeight = document.getElementsByClassName('site-head-wrapper')[0].offsetHeight,
```
hakubo: you can use - document.querySelector('').offsetHegiht

app/components/site-head.js
hakubo: probably 'reads' is better here

do we set it in anyway? or just reads it?

app/templates/components/site-head.hbs
```diff
@@ -32,13 +32,13 @@
 				isVisible=(if isSearchPage false true)
 			}}
 		</div>
+		{{#if partnerSlot}}
+			<a class="site-head__partner-logo" href="{{partnerSlot.href}}" {{action 'trackClick' partnerSlot.tracking_label preventDefault=false}}>
+				<img class="site-head__partner-logo-img" src="{{partnerSlot.image-data.url}}" alt="{{partnerSlot.title.value}}">
+			</a>
+		{{/if}}
 	</nav>
+	{{#if partnerSlot}}
```
hakubo: what is this for?
why do we render partner slot in 2 places?

app/styles/component/_site-head.scss
```diff
@@ -147,5 +159,38 @@ body:not(.no-scroll) {
...
+			position: fixed;
+		}
 	}
 }
+
+.site-head__partner-slot {
+	background: white;
+	border-bottom: 1px solid $vapor;
+	height: $partner-slot-height;
+	position: fixed;
```
hakubo: can this slot be rendered inside an element that is already fixed?
we we have only one fixed element?

app/components/category-members-grouped.js
```diff
@@ -30,7 +30,7 @@ export default Component.extend(
 
 				this.get('loadPage')(page)
 					.then(() => {
-						const navHeight = $('.site-head').outerHeight() + $('.site-head-fandom-bar').outerHeight(),
+						const navHeight = document.querySelector('site-head-wrapper').offsetHeight,
```
latata: missing `.` in selector

app/components/site-head.js
```diff
@@ -16,14 +16,14 @@ export default Component.extend(
...
-		classNameBindings: ['themeBar'],
+		classNameBindings: ['themeBar', 'partnerSlot:site-head__has-partner-slot'],
 		tagName: 'div',
 		themeBar: false,
 		closeIcon: 'close',
 		offset: 0,
 
 		defaultWikiaHomePage: 'http://fandom.wikia.com',
-
+		partnerSlot: reads('globalNavigation.partner_slot'),
```
latata: use `readOnly`

app/components/site-head.js
```diff
@@ -16,14 +16,14 @@ export default Component.extend(
 		smartBanner: service(),
 
 		classNames: ['site-head-container'],
-		classNameBindings: ['themeBar'],
+		classNameBindings: ['themeBar', 'partnerSlot:site-head__has-partner-slot'],
```
latata: I'd name the class `has-partner-slot` without `site-head__` prefiex. This is how we do it in other places.

app/styles/_wikia-variables.scss
```diff
@@ -72,9 +72,9 @@ $fandom-app-smart-banner-height: 85px;
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 $site-head-top-border-height: .375rem;
 $site-head-content-height: 3.0625rem;
-$site-head-fandom-bar-height: 1.4375rem;
+$site-head-partner-slot-height: 10px;
 $site-head-height: $site-head-top-border-height + $site-head-content-height;
-$site-head-with-fandom-bar-height: $site-head-height + $site-head-fandom-bar-height;
+$site-head-with-partner-slot-height: calc(#{$site-head-height} + #{$site-head-partner-slot-height});
```
latata: do we need to use calc here?

app/styles/component/_site-head.scss
```diff
@@ -46,14 +46,24 @@ body:not(.no-scroll) {
 	height: $site-head-height;
 }
 
+.site-head__has-partner-slot .site-head-wrapper,
+.site-head__has-partner-slot .site-head {
+	height: $site-head-with-partner-slot-height;
+}
+
+.no-scroll .site-head__has-partner-slot .site-head-wrapper,
```
latata: it looks like something we can simplify

app/styles/component/_site-head.scss
```diff
@@ -46,14 +46,24 @@ body:not(.no-scroll) {
 	height: $site-head-height;
 }
 
+.site-head__has-partner-slot .site-head-wrapper,
```
latata: it looks like something we can simplify

app/styles/component/_site-head.scss
```diff
@@ -133,6 +143,25 @@ body:not(.no-scroll) {
 	}
 }
 
+.site-head__partner-slot {
+	flex-basis: 100%;
+}
+
+.site-head__partner-logo {
+	left: 50%;
+	position: absolute;
```
latata, line 156: do we need to use absolute positioning? flexbox won't work?

app/styles/_wikia-variables.scss
rybmat: we are adding px to rem

app/styles/component/_site-head.scss
rybmat: how?

app/styles/component/_site-head.scss
rybmat: how?

app/styles/component/_site-head.scss
rybmat: flex works, but then it is too low and the whole navbar is higher than it needs to be

app/templates/components/site-head.hbs
```diff
@@ -32,13 +32,15 @@
 				isVisible=(if isSearchPage false true)
 			}}
 		</div>
+		{{#if partnerSlot}}
+			<div class="site-head__partner-slot">
+				<a class="site-head__partner-logo" href="{{partnerSlot.href}}" {{action 'trackClick'
+																						partnerSlot.tracking_label
```
hakubo: is it github or indenting here is weird?

hakubo (approved):

latata (approved):

PR #414, opened 2018-02-12
ADEN-6584 | Use ad-products@4.0.0

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-6584


package.json
```diff
@@ -31,7 +31,7 @@
   "license": "MIT",
   "dependencies": {
     "@wikia/ad-engine": "9.2.1",
```
fraszczakszymon: Could you remove this dependency please?

PR #426, opened 2018-02-15
XW-4394 | Use upstream fastboot

hakubo:


app/templates/components/fastboot-only/body-bottom.hbs
```diff
@@ -7,3 +7,5 @@
 {{/if}}
 
 {{fastboot-only/local-navigation-for-seo}}
+
+<script type="x-head-data-store" id="head-data-store">{{{data}}}</script>
```
latata, line 11: I know that shoebox probably do similar thing but are we sure `data` is safe? No XSS vulnerability?

latata (approved):

bkoval (approved):

kolek7 (approved):

PR #436, opened 2018-02-21
allow for setting 'file' queryparam

rybmat:
Correct condition to check if param is allowed

## Reviewers

@Wikia/x-wing

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -138,7 +138,7 @@ export default Controller.extend(
 			 * @returns {void}
 			 */
 			setQueryParam(name, value) {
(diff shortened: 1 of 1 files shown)


app/controllers/application.js
```diff
@@ -138,7 +138,7 @@ export default Controller.extend(
 			 * @returns {void}
 			 */
 			setQueryParam(name, value) {
-				if (name === 'file') {
+				if (name !== 'file') {
```
hakubo, line 141: :roll_eyes: sorrrrrrry :)

hakubo (approved):

PR #459, opened 2018-02-27
XW-4654 | Render images in lightbox based on data attributes instead of model

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/4654
https://github.com/Wikia/app/pull/14860/files

## Description
Rendering media thumbnails based on data attributes instead of model

## Reviewers

@Wikia/x-wing


app/components/lightbox-wrapper.js
```diff
@@ -1,26 +1,29 @@
...
 	footerHidden: false,
 	headerHidden: false,
 	header: null,
 	footer: null,
+	closeButtonHidden: false,
 
-	type: null,
-	model: null,
+	isVisible: alias('lightbox.isVisible'),
+	lightboxCloseButtonDelay: alias('lightbox.closeButtonDelay'),
```
hakubo, line 24: is this prefix lightbox needed?

hakubo: also should these be aliases?
maybe 'reads' is enough?
should this component be allowed to set these?

app/routes/application.js
```diff
@@ -197,18 +201,7 @@ export default Route.extend(
...
 				routerScroll.update();
 
-				// Because application controller needs wiki-page controller
-				// we can't be sure that media model will be ready when aplication controller is ready
 				run.scheduleOnce('afterRender', () => {
-					const file = controller.get('file'),
-						map = controller.get('map');
-
-					if (!isEmpty(file)) {
-						controller.openLightboxForMedia(file);
```
hakubo, line 208: so where this happens now?
what will open a lightbox when user opens a page with ?file=XYZ?

rybmat: https://wikia-inc.atlassian.net/browse/XW-4740

app/components/lightbox-wrapper.js
rybmat: good point, I'll check it

hakubo (approved):

app/services/lightbox.js
```diff
@@ -7,6 +7,7 @@ export default Service.extend(NoScrollMixin, {
 	isVisible: false,
 	lightboxType: null,
 	model: null,
+	preserveScrollPosition: false,
```
bkoval: What do you think about adding "noScroll = false" here to initialize it?

bkoval (approved):

app/components/article-media-thumbnail.js
```diff
@@ -22,15 +25,15 @@ export default Component.extend(
 		 */
 		itemContext: 'article',
 
-		hasFigcaption: or('caption', 'showTitle'),
+		hasFigcaption: or('model.caption', 'showTitle'),
 		isVideo: equal('type', 'video'),
 
-		isOgg: computed('mime', function () {
-			return this.get('mime') === 'application/ogg';
+		isOgg: computed('model.mime', function () {
```
bkoval: You can use a macro here to make it a bit shorter: https://guides.emberjs.com/v2.13.0/object-model/computed-properties/#toc_computed-property-macros

app/components/article-media-thumbnail.js
```diff
@@ -40,20 +43,18 @@ export default Component.extend(
 		/**
 		 * Check if image width is smaller than article container
 		 */
-		isSmall: computed('width', 'height', function () {
-			return this.get('width') <= this.get('viewportWidth');
+		isSmall: computed('model.{width,height}', function () {
```
bkoval: https://emberjs.com/api/ember/2.15/namespaces/Ember.computed/methods/gte?anchor=lte could be shorter

Also -> do we need to observe height since we only use width here?

app/components/portable-infobox.js
```diff
@@ -77,6 +78,28 @@ export default Component.extend(
 			}
 		},
 
+		click(event) {
+			const galleryActionButton = event.target.closest('.image-collection-actions button');
+
+			// switching image in pi-image-collection
+			if (galleryActionButton) {
+				const galleryWrapper = galleryActionButton.closest('.pi-image-collection');
```
bkoval: I think it might be worth it to move the logic inside if into a separate methods e.g. handleInfoboxImageCollection + add jsdoc to it. This code does a lot of custom (outside ember) DOM maneuvering so it would be beneficial to know exactly what it does. Also, we would not be cluttering the main click handler for infoboxes :)

hakubo (approved):

JamesHawking (approved):

hakubo (approved):

bkoval (approved):

JamesHawking (approved):

PR #462, opened 2018-02-28
XW-4311 | Logging anomallies in tracked url

bkoval:
This logs a situation where a tracked url is not a part of window.location.href = badly tracked url.

This is to find out:
* If the root cause to tracking problems is us not updating the url properly
* How many events are affected by this (this might be unrelated to featured video)
* On what specific urls does it happen

### DIFF
--- a/app/utils/analyzeTrackedUrl.js
+++ b/app/utils/analyzeTrackedUrl.js
@@ -11,9 +11,13 @@ export default function analyzeTrackedUrl(params) {
 		return;
 	}
 
-	const page = tracker.get('page');
+	const gaPage = tracker.get('page');
+	const actualUrl = window.location.href;
 
-	if (window.location.href.indexOf(page) === -1) {
-		logEvent('GA url does not match window.location', params);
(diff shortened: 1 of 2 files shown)


app/utils/analyzeTrackedUrl.js
```diff
@@ -0,0 +1,19 @@
...
+	const tracker = window.ga.getAll()[0];
+
+	if (!tracker) {
+		return;
+	}
+
+	const page = tracker.get('page');
+
+	if (window.location.href.indexOf(page) === -1) {
+		logEvent('GA url does not match window.location', params);
```
rybmat: @bart can we explicitly log this `page` variable and `window.location` to see the difference and compare them to what we log based on HTTP headers in event-logger?

hakubo (approved):

app/utils/analyzeTrackedUrl.js
bkoval: Great idea!

latata (approved):

rybmat (approved):

PR #472, opened 2018-03-02
XW-4679 | References do not work since the target anchor is in a collapsed section.

kolek7:
## Links

https://wikia-inc.atlassian.net/browse/XW-4679

## Description
References in articles don't work because they invariably point to a collapsed section.


app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
+				if (referencesListHead) {
+					citeNotesHeader = referencesListHead.previousElementSibling;
+				}
+			}
+
+			if (citeNotes.length && citeRefs.length) {
+
+				// If referenced section is closed, open it.
+				citeNotes.forEach((element) => {
+					element.onclick = function () {
```
hakubo: create one click() handler on whole article-content
and handle all clicks in there

app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
+						const currentCiteNote = document.getElementById(element.hash.slice(1, element.length));
+						const closest = currentCiteNote.closest('.mobile-hidden');
+
+						if (closest) {
+							let currentHeader = document.getElementById(closest.previousElementSibling.id);
+							if (currentHeader.className !== 'open-section') {
+								currentHeader.click();
```
hakubo: although this works 
it's rather ugly
you could extract body of the click handler of a header
and call it directly

app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
+
+				// If referenced section is closed, open it.
+				citeNotes.forEach((element) => {
+					element.onclick = function () {
+						const currentCiteNote = document.getElementById(element.hash.slice(1, element.length));
+						const closest = currentCiteNote.closest('.mobile-hidden');
+
+						if (closest) {
+							let currentHeader = document.getElementById(closest.previousElementSibling.id);
+							if (currentHeader.className !== 'open-section') {
```
hakubo: you can use classList API

hakubo (changes requested):

app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
 		},
 
+		/**
+		 * Handles opening sections when onclick event occurs on reference
+		 *
+		 * @returns {void}
+		 */
+		handleRefs() {
+			const citeNotes = this.element.querySelectorAll(
+					'ol[class*="references"] ' +
```
latata: do we need such a specific selector?

app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
+					'li[id^="cite_note"] ' +
+					'a[href*="cite_ref"]'
+				),
+				citeRefs = this.element.querySelectorAll('a[href*="cite_note"]'),
```
latata: I'd go with `.reference a[href^=#cite_note-]` selector

app/components/article-content.js
```diff
@@ -388,6 +390,59 @@ export default Component.extend(
...
+				referencesList = this.element.querySelectorAll('ol[class*="references"]');
```
latata: Why not `ol.references`?

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
...
 		},
 
+		/**
+		 * Handles opening sections when click event occurs on reference
+		 *
+		 * @returns {void}
+		 */
+		handleReferences(event) {
+			if (event.target.closest('a[href^="#cite_ref-"]')) {
+				const element = event.target.closest('a[href^="#cite_ref-"]'),
```
hakubo: I know we are still not consitent with that 
but per our current js style guide- https://github.com/Wikia/javascript-style-guide#variables
one const per variable

we need to finally work on our eslinter...

hakubo: also you get this element twice, could be optimized to get it just once maybe?

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
...
+					currentCiteNote = document.getElementById(element.hash.slice(1, element.length)),
+					closest = currentCiteNote.closest('section[id*="section"]');
+
+				if (closest) {
+					let currentHeader = document.getElementById(closest.previousElementSibling.id);
```
hakubo: getElementById can return null

@bkoval maybe we should write a small wrapper for common getters?

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
...
+					if (!currentHeader.classList.contains('open-section')) {
```
hakubo: you don't have to check if it contains the class
add - will add only once

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
...
+						currentHeader.classList.add('open-section');
+					}
+				}
+			}
+
+			if (event.target.closest('.reference a[href^="#cite_note-"]')) {
+				let header = this.element.querySelectorAll('ol.references');
+				if (header) {
+					header = header[0].closest('section[id*="section"]').previousElementSibling;
```
hakubo: try not to reuse variables, having two separate ones should make the code easier to understand

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
 			}
 		},
 
+		/**
+		 * Handles opening sections when click event occurs on reference
+		 *
+		 * @returns {void}
+		 */
+		handleReferences(event) {
```
hakubo: you can use object destructuring here
`{target}`
and then you can access it 'directly'

hakubo:
This is starting to look good! Thanks!

app/components/article-content.js
latata: I guess `closest.previousElementSibling === document.getElementById(closest.previousElementSibling.id)`.

latata: use `const` if you don't change variable's value.

app/components/article-content.js
```diff
@@ -388,6 +390,36 @@ export default Component.extend(
...
+				if (closest) {
+					let currentHeader = document.getElementById(closest.previousElementSibling.id);
+					if (!currentHeader.classList.contains('open-section')) {
+						currentHeader.classList.add('open-section');
+					}
+				}
+			}
+
+			if (event.target.closest('.reference a[href^="#cite_note-"]')) {
+				let header = this.element.querySelectorAll('ol.references');
```
latata: I wonder if article can have more than one references section. If yes, we should find the right section first.

app/components/article-content.js
```diff
@@ -388,6 +391,57 @@ export default Component.extend(
 			}
 		},
 
+		/**
+		 * Handles opening sections when click event occurs on CiteNote
+		 *
+		 * @returns {void}
+		 */
+		handleCiteNotes(event) {
```
hakubo: handleCiteNotes({target}) { - you can destructure arguments

app/components/article-content.js
```diff
@@ -388,6 +391,57 @@ export default Component.extend(
...
+			const {target} = event;
+
+			if (target.nodeName === 'A') {
+				if (target.hash.search('#cite_ref-') === 0) {
+					const citeNoteList = this.element.querySelectorAll(target.hash);
```
hakubo: since we always need first citeNoteList - maybe we should use querySelector?

app/components/article-content.js
```diff
@@ -388,6 +391,57 @@ export default Component.extend(
...
+					let currentSection = null;
+
+					if (citeNoteList.length === 1) {
+						currentSection = citeNoteList[0].closest('section[id*="section"]');
```
hakubo: you could 'let' here
as we don't need the var outside this if

and you could actually use const :)

app/components/article-content.js
```diff
@@ -388,6 +391,57 @@ export default Component.extend(
...
+		 */
+		handleCiteRefs(event) {
+			const {target} = event;
+
+			if (target.nodeName === 'A') {
+				if (target.hash.search('#cite_note-') === 0) {
+					const referenceListItem = this.element.querySelectorAll(target.hash);
+
+					if (referenceListItem.length === 1) {
+						const referencesSection = referenceListItem[0].parentElement.closest('section[id*="section"]');
```
hakubo: why do you need to get parentElement? can't this just run on referenceListItem?

app/components/article-content.js
```diff
@@ -388,6 +391,57 @@ export default Component.extend(
...
+						let referencesHeader = null;
+
+						if (referencesSection) {
```
hakubo: this part of code looks very similar to the code from the other function
maybe you could extract openining a section to another method? :)

app/components/article-content.js
```diff
@@ -388,6 +391,51 @@ export default Component.extend(
...
+
+		/**
+		 * Handles opening sections when click event occurs on CiteRef
+		 *
+		 * @returns {void}
+		 */
+		handleCiteRefs({target}) {
+			if (target.nodeName === 'A') {
+				if (target.hash.search('#cite_note-') === 0) {
+					const referenceListItem = this.element.querySelector(target.hash);
```
bkoval: Make sure to leave one blank line after declarations of consts / lets 🚓 🕺

app/components/article-content.js
```diff
@@ -388,6 +391,51 @@ export default Component.extend(
...
+		},
+
+		/**
+		 * Handles opening sections when click event occurs on CiteNote
+		 *
+		 * @returns {void}
+		 */
+		handleCiteNotes({target}) {
+			if (target.nodeName === 'A') {
+				if (target.hash.search('#cite_ref-') === 0) {
```
bkoval: How about combining these two if's into one? This way we could gain some readability as there would be less indentations:

```
if (target.nodeName === 'A' && target.hash.search('#cite_ref-') === 0) {
```

app/components/article-content.js
```diff
@@ -388,6 +391,51 @@ export default Component.extend(
...
+		 * Opens a parent section of passed element if it's closed
+		 *
+		 * @returns {void}
+		 */
+		openSection(element) {
+			if (element) {
+				const section = element.closest('section[id*="section"]');
+				if (section) {
+					const header = section.previousElementSibling;
+					if (header) {
```
bkoval: What if the ref is in the "zero section" (the article section that starts directly below the article title and ends before the first H2). This section is always expanded, even on mobile.

I would consider checking if tagName === 'H2' before appending the 'open-section' class. What do you think?

app/components/article-content.js
```diff
@@ -388,6 +391,51 @@ export default Component.extend(
...
+						header.classList.add('open-section');
```
bkoval: What about scrolling the ref into view after expanding the section? Does that work now based on url hash or should we think of adding some JS to scroll the given ref into view?

bkoval (changes requested):
Great job! A few minor comments but overall it looks fine!

app/components/article-content.js
```diff
@@ -388,6 +390,50 @@ export default Component.extend(
 			}
 		},
 
+		/**
+		 * Opens a parent section of passed element if it's closed
+		 *
+		 * @returns {void}
+		 */
+		openSection(element) {
+			if (element) {
```
xkxd: this check is redundant. if "element" will be null/undefined it'll throw a TypeError in 
```const offsetY = reference.getBoundingClientRect().top + window.scrollY;```
anyway

\+ please add ```@param {Element} element``` to jsdocs

File: app/components/article-content.js [hunk 23]
```diff
@@ -388,6 +390,50 @@ export default Component.extend(
...
+						header.classList.add('open-section');
+					}
+				}
+			}
+		},
+
+		/**
+		 * Handles opening sections when click event occurs on references
+		 *
+		 * @returns {void}
```
xkxd: please add `@param {MouseEvent} event` to jsdocs

xkxd (approved):

bkoval (approved):

PR #478, opened 2018-03-06
XW-4740 | open lightbox when opening article with file query param

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/XW-4740

@Wikia/x-wing


JamesHawking (approved):

app/components/lightbox-media.js
```diff
@@ -177,6 +179,7 @@ export default Component.extend(
 		updateState() {
 			this.updateHeader();
 			this.updateFooter();
+			this.get('lightbox').setQueryParam(normalizeToUnderscore(this.get('currentMedia.title')));
```
latata: use `set('file', ...)`

PR #489, opened 2018-03-13
 ADEN-6847 | Start loading ads faster

kvas-damian:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-6847

## Description

Currently, ads code starts loading after Ember is initialized and ready. It means that download time additionally delays ads requests and impressions. According to requests waterfall, the browser isn't loading anything while Ember code is being executed. This PR changes when we start loading ads code. Now we start the request immediately after InstantGlobals are loaded.

Before the change:
zaznaczenie_282

After the change:
zaznaczenie_285

## Reviewers

@Wikia/adeng @hakubo


app/inline-scripts/load-ads.js
```diff
@@ -0,0 +1,44 @@
...
+			wikiVariables.cdnRootUrl + '/__am/' + wikiVariables.cacheBuster + '/groups/-/mercury_ads_js',
+			true,
+			function () {
+				window.document.dispatchEvent(new Event('adsScriptLoaded'));
+			}
+		);
+	});
+
+	window.waitForAds = function (callback) {
+		function onAdsLoaded() {
```
hakubo: this will create as many functions as calls to waitForAds - intentional?

kvas-damian: Nice catch. Fixed and also simplified the code a little bit.

app/inline-scripts/load-ads.js
```diff
@@ -0,0 +1,36 @@
...
+
+		getterAdsQueue.forEach(function (queuedCallback) {
+			queuedCallback();
+		});
+
+		getterAdsQueue = [];
+	}
+
+	window.getInstantGlobal('wgSitewideDisableAdsOnMercury', function (wgSitewideDisableAdsOnMercury) {
+		if (wgSitewideDisableAdsOnMercury || (new URL(document.location)).searchParams.get('noexternals')) {
```
fraszczakszymon, line 16: `noexternals` is handled in https://github.com/Wikia/mobile-wiki/pull/489/files#diff-2168f026d657db1526ae92392f844a3fR91. Isn't it enough?

kvas-damian: It's required to avoid loading ads code when `noexternals` is present.

fraszczakszymon (approved):

PR #491, opened 2018-03-14
XW-4785 | get internal tracking data from shoebox instead of head data store

rybmat:
## Links

* https://wikia-inc.atlassian.net/browse/XW-4785

## Description

articleId and namespace was not set in data-head-store, thus undefined was sent for these fields to internal datawarehouse. This is temporary fix, ticket will be created to cleanup.

## Reviewers

@Wikia/x-wing

### DIFF
--- a/app/mixins/wiki-page-handler.js
+++ b/app/mixins/wiki-page-handler.js
@@ -95,10 +95,9 @@ export default Mixin.create({
 						}
 
 						shoebox.put('wikiPage', dataForShoebox);
-						this.get('simpleStore').setProperties({
-							namespace: get(dataForShoebox, 'data.ns'),
-							articleId: get(dataForShoebox, 'data.details.id'),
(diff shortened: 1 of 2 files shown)


app/mixins/wiki-page-handler.js
```diff
@@ -95,10 +95,9 @@ export default Mixin.create({
 						}
 
 						shoebox.put('wikiPage', dataForShoebox);
-						this.get('simpleStore').setProperties({
-							namespace: get(dataForShoebox, 'data.ns'),
-							articleId: get(dataForShoebox, 'data.details.id'),
-							isMainPage: get(dataForShoebox, 'data.isMainPage')
+						shoebox.put('trackingData', {
+							articleId: dataForShoebox.data.details.id,
```
hakubo: use get()

bkoval (approved):

hakubo (approved):

PR #493, opened 2018-03-14
XW-4743 | Fandom App webview events tracking added.

kolek7:
## Links

https://wikia-inc.atlassian.net/browse/XW-4743

## Reviewers

@Wikia/x-wing


app/utils/track.js
```diff
@@ -120,9 +120,12 @@ function isPageView(category) {
  * @returns {void}
  */
 export function track(params) {
+	const isFandomApp = new URL(window.location).searchParams.get('mobile-app');
+	const appOrMercury = (isFandomApp === 'true' ? 'fandom-app-' : 'mercury-');
```
rybmat: can we name this const differently? `categoryPrefix` for example?

app/utils/track.js
```diff
@@ -120,9 +120,12 @@ function isPageView(category) {
...
+
 	const trackingMethod = params.trackingMethod || 'both',
 		action = params.action,
-		category = params.category ? `mercury-${params.category}` : null,
+		category = params.category ? appOrMercury + params.category : null,
```
rybmat: we can use string templates here `${categoryPrefix}-${params.category}`
(note: remove trailing `-` from the categoryPrefix const)

rybmat (changes requested):

hakubo (approved):

app/utils/track.js
```diff
@@ -120,9 +120,12 @@ function isPageView(category) {
  * @returns {void}
  */
 export function track(params) {
+	const isFandomApp = new URL(window.location).searchParams.get('mobile-app') === 'true';
+	const trackingCategoryPrefix = (isFandomApp ? 'fandom-app' : 'mercury');
+
 	const trackingMethod = params.trackingMethod || 'both',
```
hakubo: while you're at it could you split to const per variable?

app/utils/track.js
```diff
@@ -120,9 +120,12 @@ function isPageView(category) {
  * @returns {void}
  */
 export function track(params) {
+	const isFandomApp = new URL(window.location).searchParams.get('mobile-app') === 'true';
```
hakubo: this function runs also on Fastboot
on  line 133 and 137 there are neede checks - this should probably happen after these lines

xkxd: i'd suggest moving those fastboot/window checks at the top of `track` method

rybmat (approved):

hakubo (approved):

PR #513, opened 2018-03-21
XW | use greenkeeper-lockfile to manage package-lock.json

hakubo:
https://github.com/greenkeeperio/greenkeeper-lockfile

This should make updating dependencies much easier

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -88,26 +88,6 @@
         "samsam": "1.3.0"
       }
     },
-    "@wikia/ad-engine": {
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -4342,13 +4342,6 @@
         "minimalistic-assert": "1.0.0"
       }
     },
-    "design-system": {
```
latata: What happened with DS?

latata (approved):

package-lock.json
hakubo: it's installed via github not registry,
probably this
I'll downgrade npm on my laptop and see what happens

PR #528, opened 2018-03-22
XW-2275 | call DS API via community domain and default api url

latata:
https://wikia-inc.atlassian.net/browse/XW-2275
@Wikia/x-wing

Wait with merging 24hours after releasing this: https://github.com/Wikia/app/pull/15047


JamesHawking (approved):

app/models/navigation.js
```diff
@@ -7,8 +7,8 @@ export default EmberObject.extend({
 	fetchAll(host, wikiId, language) {
 		const url = buildUrl({
 			host,
-			path: `/api/v1/design-system/wikis/${wikiId}/${language}/`,
-			wiki: 'www'
+			/* eslint-disable max-len */
```
hakubo: instead of disabling eslint here
what do you think about
`/wikia.php?controller=DesignSystemApi&method=getAllElements\
&product=wikis&id=${wikiId}&lang=${language}`

latata: IMO with disabling eslint the path is more readable and it's at least 2 bytes less code ;)

hakubo: 1. maybe :)
2. nope, https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBCELwQBgGwPaIgHQFADtkAcBTeAbiA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&lineWrap=true&presets=es2015%2Ces2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2 :)

hakubo (approved):

PR #537, opened 2018-03-23
XF-70 | stop using observer

hakubo:
Bunch of fixes along with it
- stop using observer
- fix for horizontal scrolling
- fix search suggestion click


app/inline-scripts/tracking-internal.js
```diff
@@ -3,10 +3,14 @@
 		return;
 	}
 
-	const shoeboxTrackingData = JSON.parse(document.querySelector('#shoebox-trackingData').innerHTML);
+	const trackingData = document.querySelector('#shoebox-trackingData');
```
rybmat: IMO this should be named `trackingDataShoebox` and the one below just `trackingData`

app/components/article-content.js
```diff
@@ -40,59 +40,49 @@ export default Component.extend(
 		dir: reads('wikiVariables.language.contentDir'),
 		isFastBoot: reads('fastboot.isFastBoot'),
```
rybmat: is this used somewhere? Asking because you deleted the usage of this field in .hbs

rybmat (approved):

rybmat (approved):

bkoval (approved):

PR #538, opened 2018-03-23
Code coverage

hakubo:


rybmat (approved):

xkxd:
what is the actual use case of this? will this e.g. block PR from merging in case of decreased code coverage?

hakubo:
@xkxd sounds good. Do you think it would be helpful?
for now I was thinking of just 'knowing' what is the code coverage

rybmat:
blocking PR just because of decreased code coverage does not sound good, it will be pain in the ass in case of hot fixes for P1/P2 tickets. But I'm fine with experience like with outdated branch: 
screen shot 2018-03-26 at 11 24 23

xkxd:
@hakubo it'll be good to have such condition because in another case I'd assume it'll be a constant downtrend, but as we spoke there is a lot of possibilities to improve this in the future.

@rybmat yes, that's what I meant - you'd still be able to merge it as an admin, it'll just notice you that some jenkins checks (i.e. code coverage (CC)) failed.

and I guess there is a possibility that it will fail when CC will decrease by more than eg. 1-2% to avoid writing tests for one-liners etc.

PR #542, opened 2018-03-26
ADEN-6313 Add setting window.wgContentLanguage

dmnsobczak:


app/mixins/ads.js
```diff
@@ -26,6 +26,9 @@ export default Mixin.create({
 	init() {
 		this._super(...arguments);
 		this.renderAdComponent = getRenderComponentFor(this);
+
+		// Set proper UAP language
+		window.wgContentLanguage = this.get('wikiVariables.language.content');
```
fraszczakszymon, line 31: Maybe `context.set('options.contentLanguage')`?

fraszczakszymon: Actually, it would be enough to set this in the bridge (app repo) and close this pull request.

dmnsobczak: I think you right. Closing.

PR #549, opened 2018-03-27
XF-47 | Upgrade Ember to 3.0.

kolek7:
## Links
https://wikia-inc.atlassian.net/browse/XF-47

## Reviewers
@Wikia/x-wing


Dockerfile.prod
```diff
@@ -19,6 +19,10 @@ RUN addgroup -S mobile-wiki && \
 # install missing stuff
     apk add --no-cache --virtual .gyp python make g++ git && \
 
+# related to upgrading ember to 3.0
+    npm run nuke
+    TODO: remove later
```
bkoval: Could you add a ticket number (the one to remove the  nuke)?

rybmat: Shouldn't TODO be commented?

rybmat (changes requested):

config/targets.js
```diff
@@ -1,3 +1,18 @@
+'use strict';
+
+const browsers = [
+	'last 1 Chrome versions',
+	'last 1 Firefox versions',
+	'last 1 Safari versions'
```
bkoval: Since chrome and firefox have auto-update feature it is safe to use only one version, but with Safari I'd check our compatibility table:

http://community.wikia.com/wiki/Help:Supported_browsers

iOS9 is the last supported so let's make sure Safari for this version of iOS is also supported here.

Dockerfile.prod
kolek7: Right...

config/targets.js
```diff
@@ -1,3 +1,18 @@
...
+];
+
+const isCI = !!process.env.CI;
+const isProduction = process.env.EMBER_ENV === 'production';
+
+if (isCI || isProduction) {
```
bkoval: We do not support IE11 on mobile-wiki, we can safely remove the support.

bkoval (changes requested):

Dockerfile.prod
```diff
@@ -19,6 +19,10 @@ RUN addgroup -S mobile-wiki && \
 # install missing stuff
     apk add --no-cache --virtual .gyp python make g++ git && \
 
+# related to upgrading ember to 3.0
+    npm run nuke
```
xkxd: ❤️

rybmat (approved):

config/targets.js
```diff
@@ -1,3 +1,18 @@
+'use strict';
+
+const browsers = [
```
xkxd: where is it used? you're still exporting different object

xkxd (changes requested):

config/targets.js
```diff
@@ -1,3 +1,18 @@
...
+	'last 1 Chrome versions',
```
xkxd: `ChromeAndroid` would be better

config/targets.js
```diff
@@ -1,3 +1,18 @@
...
+	'last 1 Firefox versions',
```
xkxd: as well as `FirefoxAndroid`

bkoval (approved):

package.json
```diff
@@ -21,7 +21,8 @@
     "setup": "npm install",
     "svgo": "bash ./tasks/optimize-svg.sh",
     "lint:js": "eslint ./*.js app config lib server tests",
-    "test": "ember test"
+    "test": "ember test",
+    "nuke": "npm cache clean --force && rm -rf node_modules/ && rm package-lock.json"
```
xkxd, line 25: is it still needed?

hakubo (approved):

xkxd (approved):

package.json
kolek7: Yeah, it's a script that we'll be using in the future

hakubo (approved):

package.json
xkxd: then why don't you call `npm run nuke` in dockerfile?

xkxd: then why don't you call `npm run nuke` in dockerfile?

xkxd (changes requested):

package.json
kolek7: It was causing an error when deploying to sandbox.
Considering that we'll be removing this code next week anyway I don't really think that this matters in the long run.

xkxd (approved):

PR #563, opened 2018-03-29
XF-9 | LI isDestroyed check

latata:
https://wikia-inc.atlassian.net/browse/XF-9
@Wikia/x-wing


hakubo (approved):

app/components/recirculation-prefooter.js
```diff
@@ -123,11 +123,13 @@ export default Component.extend(
 						}));
 
 					run.scheduleOnce('afterRender', () => {
```
hakubo, line 125: why do we even have this afterRender here?

latata: I guess this is because we need to wait for recirc items to render to setup tracking

rybmat (approved):

hakubo (approved):

PR #588, opened 2018-04-04
XW-4829 | is destroyed check for destroyadslotcomponents

latata:
https://wikia-inc.atlassian.net/browse/XW-4829
@Wikia/x-wing


hakubo (approved):

latata:
Let's merge it just to current release and verify if it helps.

PR #599, opened 2018-04-05
XW-4847 | Fix abtest tracking

bkoval:
@Wikia/x-wing 

Currently a race condition causes GA Custom Dimension for ABTest to not be sent. We need to wait for AB Test load with initializing MW


app/index.html
```diff
@@ -35,8 +35,8 @@
 	})(window.M);
 </script>
 
-{{content-for 'instant-globals'}}
-<script src="/__load/-/only=scripts/wikia.ext.instantGlobals,instantGlobalsOverride,abtesting,abtest" onload="onInstantGlobalsLoaded()" onerror="onInstantGlobalsError()" defer></script>
+{{content-for 'async-scripts'}}
```
hakubo: how about async-scripts-handlers
async-scrips - sounds to me like this will add some async scripts

JamesHawking (approved):

vendor/inline-scripts/mediawiki-scripts-handlers.js
```diff
@@ -0,0 +1,53 @@
+(function () {
+	var gettersQueue = [];
+	var callbacksQueue = [];
+	var isListening = false;
+
+	window.onAsyncScriptsLoaded = function () {
```
rogatty, line 6: How about using `window.M` instead of polluting the global namespace?

bkoval: Do you think it's safe to set properties on window.M outside Mercury-shared?

rogatty: Huh, I thought we already do this but we don't. This could be a bad practice, I'm taking it back :slightly_smiling_face:

rybmat (approved):

latata (approved):

hakubo (approved):

PR #608, opened 2018-04-06
Turn on hbs minifier

hakubo:
We had 2 places where we relied on quircks of HTML/CSS that broke some styling when hbs minifier got introduced

both of these places where on main page and the fix was to use display flex

main page of muppet 3.2kb less


app/styles/component/_curated-content.scss
```diff
@@ -6,28 +6,21 @@
 	padding: 0 $article-horizontal-padding / 3;
 
 	.curated-content-items {
-		line-height: 0;
-		margin: 0 auto;
-		text-align: justify;
+		display: flex;
+		flex-wrap: wrap;
+		justify-content: space-evenly;
```
latata: It's not supported by mobile browsers. Can't we use `space-between` here?

app/styles/component/_curated-content.scss
```diff
@@ -6,28 +6,21 @@
...
 
 		// this pseudoelement lets the last inline item to be justified to the right if it's even
 		&::after {
 			content: '';
-			display: inline-block;
-			width: 100%;
+			width: calc(50% - 8px);
```
latata, line 16: `8px` - magic number?

app/styles/component/_curated-content.scss
```diff
@@ -6,28 +6,21 @@
...
-		line-height: initial;
-		margin-bottom: $curated-content-image-spacing;
-		vertical-align: top;
-		width: calc(50% - #{$curated-content-image-spacing / 2});
-
-		@media #{$wds-breakpoint-xl-up} {
-			width: calc(33.33% - #{$curated-content-image-spacing / 2});
-		}
+		margin-bottom: 5px;
+		max-width: 300px;
```
latata, line 22: Why do we need `max-width` here?

package-lock.json
```diff
@@ -4,6 +4,272 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
```
latata: It looks like too much changes here? Why is that?

app/styles/component/_curated-content.scss
hakubo: so it at above ~600px we brake it into 3 columns

app/styles/component/_curated-content.scss
hakubo: yup, margins and such

latata: can't we calculate it with some variables?

bkoval (approved):

PR #612, opened 2018-04-09
XW-4645 | Implement Mobile Site link on Oasis for mobile

kolek7:
## Links
https://wikia-inc.atlassian.net/browse/XW-4645

## Description
Removing 'useskin=oasis' from the link because it's not needed, cookie is set so it works fully without it.
Also removed unused css class.

## Reviewers
@Wikia/x-wing


app/templates/components/global-footer/global-footer-bottom-bar.hbs
```diff
@@ -13,8 +13,8 @@
 			ns='design-system'
 		}}}
 	</div>
-	<div class="wds-global-footer__bottom-bar-row wds-has-border-top global-footer-full-site-link-wrapper">
-		<a {{action 'fullSiteClicked' preventDefault=false}} href="?useskin=oasis" rel="nofollow" class="wds-global-footer__button-link">
+	<div class="wds-global-footer__bottom-bar-row wds-has-border-top">
+		<a {{action 'fullSiteClicked' preventDefault=false}} href rel="nofollow" class="wds-global-footer__button-link">
```
rybmat: is empty `href` valid?

kolek7: Yeah, it's valid, this will refresh the page on click now.

bkoval: Let's not use an anchor tag if it's not used as a link. FOr such onclick actions I propose to use a button or at least a div with aria "role=button". For better crawlability. You can read more about it here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role

bkoval (changes requested):

rybmat (approved):

JamesHawking (approved):

tests/acceptance/footer-test.js
```diff
@@ -24,7 +24,7 @@ module('Acceptance | Footer', (hooks) => {
 			.hasText('global-footer-fandom-overview-link-vertical-movies');
 		assert.dom('.wds-is-tv').hasTextContaining('tv');
 		// assert.dom('.wds-global-footer__bottom-bar-row a[href*="Licensing"]').exists();
-		assert.dom('a[href="?useskin=oasis"]').hasText('global-footer-full-site-link');
+		assert.dom('div[role="button"].wds_is_text').hasText('global-footer-full-site-link');
```
bkoval: I think what you want to use is ".wds-is-text" no ".wds_is_text" :)

app/templates/components/global-footer/global-footer-bottom-bar.hbs
```diff
@@ -13,9 +13,10 @@
 			ns='design-system'
 		}}}
 	</div>
-	<div class="wds-global-footer__bottom-bar-row wds-has-border-top global-footer-full-site-link-wrapper">
```
bkoval: If we'd leave the "global-footer-full-site-link-wrapper" on a wrapper, then we could have a much more specific selector in acceptance tests. What do you think?

Also, did you remove it because it wasn't used anywhere?

bkoval (changes requested):

app/templates/components/global-footer/global-footer-bottom-bar.hbs
kolek7: Yeah, it wasn't used anywhere else.

bkoval (approved):

PR #618, opened 2018-04-11
XF-368 | Language path support

latata:
https://wikia-inc.atlassian.net/browse/XF-368
@Wikia/x-wing


rybmat (approved):

app/components/fastboot-only/body-bottom.js
```diff
@@ -47,6 +49,13 @@ export default Component.extend({
 		}, simpleStore));
 	}),
 
+	asyncScriptsPath: computed(function () {
+		const langPath = getLanguageCodeFromRequest(this.get('fastboot.request')),
```
hakubo: one const per var -
should we enable this in eslint?
maybe it can fix it automatically?

app/components/user-avatar.js
```diff
@@ -19,7 +21,10 @@ export default Component.extend({
 	 * @returns {string}
 	 */
 	profileUrl: computed('profileName', function () {
+		const langPath = getLanguageCodeFromRequest(this.get('fastboot.request'));
+
 		return buildUrl({
+			langPath,
```
hakubo: langPath: getLanguageCodeFromRequest(this.get('fastboot.request'))

less code, same readability IMO

app/components/wikia-search.js
```diff
@@ -205,6 +207,7 @@ export default Component.extend(
 		getSearchURI(phrase) {
 			return buildUrl({
 				host: this.get('wikiVariables.host'),
+				langPath: getLanguageCodeFromRequest(this.get('fastboot.request')),
```
hakubo: here you actually do this - I'd keep this style

app/models/wikia-in-your-lang.js
```diff
@@ -43,7 +45,8 @@ export default EmberObject.extend(LanguagesMixin, {
 	 */
 	load() {
 		const browserLang = this.getBrowserLanguage(),
-			model = getFromCache(browserLang);
+			model = getFromCache(browserLang),
```
hakubo: one const per variable

app/router.js
```diff
@@ -1,27 +1,51 @@
...
 import RouterScroll from 'ember-router-scroll';
 import config from './config/environment';
+import getLanguageCodeFromRequest from './utils/language';
 
 const Router = EmberRouter.extend(RouterScroll, {
 	location: config.locationType,
-	rootURL: config.rootURL
+	rootURL: config.rootURL,
+
+	_buildDSL() {
```
hakubo: please document why we need this

app/utils/article-link.js
```diff
@@ -47,7 +47,7 @@ export default function getLinkInfo(basePath, title, hash, uri, queryString) {
 			 *     3: "Kermit_the_Frog"
 			 *     4: "#Kermit_on_Sesame_Street"
 			 */
-			article = local.match(/^(\/(wiki))\/([^#]+)(#.*)?$/);
+			article = local.match(/^(\/[a-z]{2,3}(?:-[a-z-]{2,12})?)?(\/(wiki))\/([^#]+)(#.*)?$/);
```
hakubo: should example above be updated?

app/utils/language.js
```diff
@@ -0,0 +1,13 @@
+import config from '../config/environment';
```
hakubo: both not used?

hakubo (changes requested):
what do you think about writing some tests for it to make sure we support both variations?

app/utils/language.js
```diff
@@ -0,0 +1,13 @@
...
+import {escapeRegex} from './string';
+
+/**
+ * @param {Object} request - FastBoot request
+ * @returns {string|null}
+ */
+export default function getLanguageCodeFromRequest(request = null) {
+	const path = request ? request.get('path') : window.location.pathname,
+		matches = path.match(/^(\/[a-z]{2,3}(?:-[a-z-]{2,12})?)/);
```
rogatty: This pattern is used in 3 places. Can we reuse it using e.g. `new RegExp(langPathPart + 'the rest of pattern')`?

app/components/fastboot-only/body-bottom.js
latata: I can create a ticket to enable linter and fix it.

app/models/widget-discussions.js
```diff
@@ -1,9 +1,9 @@
 import {inject as service} from '@ember/service';
-import EmberObject, {getWithDefault, get} from '@ember/object';
```
hakubo: useless computed import?

app/utils/language.js
```diff
@@ -0,0 +1,12 @@
+export const langPathRegexp = '(/[a-z]{2,3}(?:-[a-z-]{2,12})?)';
+
+/**
+ * @param {Object} request - FastBoot request
+ * @returns {string|null}
+ */
+export default function getLanguageCodeFromRequest(request = null) {
```
hakubo: if I don't pass request, it'll fail on fastboot
also it won't be taking it from request so it might be confusing

I'd propose to pass path to the function (getLanguageCodeFromPath), following least amount of power rule
then behaviour can be more deterministic

also testing this fuction would be easy
oh yeah, can we write a simple unit test for it?

app/utils/article-link.js
```diff
@@ -47,7 +48,7 @@ export default function getLinkInfo(basePath, title, hash, uri, queryString) {
 			 *     3: "Kermit_the_Frog"
 			 *     4: "#Kermit_on_Sesame_Street"
```
rogatty: I think that's outdated.

app/utils/article-link.js
```diff
@@ -61,9 +62,9 @@ export default function getLinkInfo(basePath, title, hash, uri, queryString) {
```
rogatty: It's not shown in the diff but links to the main page might be broken, I think. They lack the `langPath` part.

rogatty: Fixed in https://github.com/Wikia/mobile-wiki/pull/618/commits/e8d26be07056d57729957f528ae7285e6212f422

rogatty (approved):

app/components/fastboot-only/body-bottom.js
```diff
@@ -48,6 +50,13 @@ export default Component.extend({
 		}, simpleStore));
 	}),
 
+	asyncScriptsPath: computed(function () {
```
hakubo, line 52: what will happen if someone goes from glee.wikia.com/de to glee.wikia.com/pl?
I think we should reload a page right?

rogatty: I added a test case to make sure it happens in https://github.com/Wikia/mobile-wiki/pull/618/commits/23fecfede869717bebada9f3020202837806c2b2

app/models/search.js
```diff
@@ -15,7 +15,9 @@ export default EmberObject.extend({
 	totalItems: 0,
 	totalBatches: 0,
 	wikiVariables: service(),
+	fastboot: service(),
```
latata: It's not needed anymore.

package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "rm -rf node_modules/ dist/ tmp/",
     "crowdin": "bash ./tasks/download-translations.sh",
     "dev": "ember serve",
-    "fastboot-debug": "ember build && node --inspect --debug-brk ./node_modules/.bin/ember serve --no-build",
+    "fastboot-debug": "bash ./fastboot-debug.sh",
```
hakubo: you removed build from it, intentional?

tests/acceptance/lang-path-test.js
```diff
@@ -0,0 +1,28 @@
...
+	setupApplicationTest(hooks);
+
+	hooks.beforeEach(function () {
+		mockFastbootService(this.owner);
+		this.owner.register('service:wikiUrls', wikiUrlsServiceStub);
+	});
+
+	test('visiting /pl/wiki/test', async (assert) => {
+		const result = await visit('/pl/wiki/test');
+		assert.ok(result);
```
hakubo, line 21: how about asserting if currentURL is the one that is expected?

assert.ok - can assert on promises? neat!

hakubo (approved):

app/models/search.js
rogatty: Removed :+1:

package.json
rogatty: It was a mistake, thanks for the good catch. Re-added in https://github.com/Wikia/mobile-wiki/pull/618/commits/3b5f4f9786e0eb7b4d4d0544276ddea2302d8d20

tests/acceptance/lang-path-test.js
rogatty: Good idea, added in https://github.com/Wikia/mobile-wiki/pull/618/commits/0193b4105653b8ec5ee4fc135651e246621ccc66

fastboot-server/middlewares.js
```diff
@@ -31,8 +31,8 @@ module.exports = {
 		 * Fastboot doesn't support POST requests so we rewrite them on express to GET
 		 * Additionally we have to enable POST body parser for this route to get data that was posted
 		 */
-		app.use('/article-preview', bodyParser.urlencoded({extended: true, limit: '10mb'}));
-		app.use('/article-preview', methodOverride(() => {
+		app.use(/^(\/[a-z]{2,3}(?:-[a-z-]{2,12})?)?\/article-preview/, bodyParser.urlencoded({extended: true, limit: '10mb'}));
```
hakubo: this regexp is slowly creeping in into my night dreams...

hakubo (approved):

hakubo (approved):

PR #621, opened 2018-04-12
ADEN-7004 Adds AdEngine assets

leshniak:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7004

## Description

Adds AdEngine assets optionally and asynchronously.

## Reviewers

@hakubo  @fraszczakszymon


app/modules/adengine-assets.js
```diff
@@ -0,0 +1,61 @@
...
+			styles.rel = 'stylesheet';
+			styles.href = assetUrls.adProductsStyles;
+			document.head.appendChild(styles);
+
+			this.wasStyleLoadInitialized = true;
+		}
+	}
+
+	loadScript(url) {
+		return new Promise((resolve) => {
```
hakubo: we also have $script to load scripts which is I think more robust

app/modules/adengine-assets.js
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	load() {
+		this.loadStyles();
+
+		return this.loadScripts();
+	}
+}
+
+export default new AdEngineAssets();
```
hakubo: not sure if exporting an instance is a good practice
I'd say if you need this as a singleton - just use module - you don't need class for that

app/modules/adengine-assets.js
```diff
@@ -0,0 +1,61 @@
+import {Promise} from 'rsvp';
+
+export const assetUrls = {
+	adEngineScript: '/mobile-wiki/assets/adengine/ad-engine.global.js',
+	adProductsScript: '/mobile-wiki/assets/adengine/ad-products.global.js',
+	geoScript: '/mobile-wiki/assets/adengine/geo.global.js',
+	adProductsStyles: '/mobile-wiki/assets/adengine/ad-products.css'
```
fraszczakszymon: We already have styles added:
https://github.com/Wikia/mobile-wiki/blob/d22c6623abd692fe45453721c4730d4d6a653f72/app/styles/module/_ads.scss#L4

We can skip this part in this place

leshniak: huh, ofc :)

fraszczakszymon:
@leshniak can you revert artifactory changes from package-lock.json?

leshniak:
@fraszczakszymon done

dmnsobczak (approved):
LGTM. Make sure to (probably manually) reapply some new changes do dev branch since last merge (and fix tests ofc).

PR #624, opened 2018-04-12
XW-4752 | Prevent font swap for repeated view.

JamesHawking:
## Links

* http://wikia-inc.atlassian.net/browse/XW-4752


app/index.html
```diff
@@ -18,9 +18,14 @@
 	<i></i>
 </div>
 <script>
-	document.fonts && document.fonts.ready && document.fonts.ready.then(function () {
+	if (localStorage.getItem('hasRubikFont')) {
```
rybmat: someone (@bkoval or @latata) said that cookie would be better for some reason. Don't remember the exact reason but I remember that it seemed to be a valid argument for me. Please consult it

rybmat (changes requested):

app/index.html
hakubo: also localStorage can throw if user is in incognito mode
also this means that if a browser chose to invalidate cache
we will show blank page to a user for a bit

hakubo: on yet another hand - google caches this fonts for pretty much eternity
"public, max-age=31536000"

rybmat (approved):

hakubo (approved):

PR #654, opened 2018-04-23
ADEN-7019 Uncollapse section on mobile per geo

dmnsobczak:


app/components/article-content.js
```diff
@@ -534,6 +538,10 @@ export default Component.extend(
 		handleCollapsibleSections() {
 			toArray(this.element.querySelectorAll('h2[section]'))
 				.forEach((header) => header.addEventListener('click', this.handleCollapsibleSectionHeaderClick.bind(this)));
+		},
+
+		uncollapseSections() {
+			toArray(this.element.querySelectorAll('h2[section]')).forEach((header) => header.classList.add('open-section'));
```
hakubo: you'll be missing calling this function
https://github.com/Wikia/mobile-wiki/pull/654/files#diff-93fe8a4709c07d310d26b67237a9cb5bR529
some things might not work as expected.

also I assume you're fine with jumping content, yes? I mean user will load an article and then after some time it'll expand.

dmnsobczak: Thanks for the advice, I made some changes.
By default, wgVariable will be set to XX and for testing and analysing purposes on small percentage of users we're fine with that.

app/components/article-content.js
```diff
@@ -534,6 +543,15 @@ export default Component.extend(
 		handleCollapsibleSections() {
 			toArray(this.element.querySelectorAll('h2[section]'))
 				.forEach((header) => header.addEventListener('click', this.handleCollapsibleSectionHeaderClick.bind(this)));
+		},
+
+		uncollapseSections() {
+			toArray(this.element.querySelectorAll('h2[section]'))
+				.forEach((header) => {
+					if (!header.classList.contains('open-section')) {
```
v3rth: It could be removed by adding to h2 selector `:not(.open-section)`. I'm not sure about performance, but it should be cleaner

dmnsobczak: You right, cleaner and that should be faster too.

nandy-andy (approved):

app/modules/ads.js
```diff
@@ -364,6 +364,10 @@ class Ads {
 		return !!document.querySelector('.wds-global-footer');
 	}
 
+	isArticleSectionCollapsed() {
+		return this.adsContext && this.adsContext.opts && this.adsContext.opts.mobileSectionsCollapse;
```
v3rth: You can reach any variable in adsContext using `get`, `this.adsContext('opts.mobileSectionsCollapse')` should work

v3rth (approved):

v3rth (approved):

v3rth:
Merge after app [release-651](https://github.com/Wikia/app/releases) cutoff

PR #660, opened 2018-04-25
SUS-4529 | render <math> on the client-side in Mercury

macbre:
## Links

* http://wikia-inc.atlassian.net/browse/SUS-4529
* http://macbre.macbre.wikia-dev.pl/wiki/Math

## Description

As we're migrating our MediaWiki app to Kubernetes we need to prepare a Docker image to handle all features that we currently support. This includes rendering TeX syntax in `<math>` tags. The current solution uses server-side rendering that requires heavy dependencies (ghostscript, latex, imagemagick) that weight ~2 GB (!).

More up to date version of MediaWiki extension that handles `<math>` tags makes use of MathJax JS library that performs rendering on the client-side. This PR introduces this feature in `mobile-wiki` - see https://github.com/Wikia/app/pull/15159 for desktop skin changes.

## QA notes

This PR needs to be tested together with `app@SUS-4529-math-rendering` branch deployed on a sandbox.

## Screenshots

math macbre wiki fandom powered by wikia

math macbre wiki fandom powered by wikia


rogatty:
Can you instead follow the widget pattern we have? See:

- https://github.com/Wikia/app/blob/dev/extensions/wikia/ApesterTag/ApesterTagController.class.php
- https://github.com/Wikia/mobile-wiki/blob/dev/app/components/article-content.js#L352-L415
- https://github.com/Wikia/mobile-wiki/blob/dev/app/templates/components/widget-apester.hbs
- https://github.com/Wikia/mobile-wiki/blob/dev/app/components/widget-apester.js
- https://github.com/Wikia/mobile-wiki/blob/dev/app/mixins/widget-script-state.js

Thanks to that:
- we won't need to manually query the DOM (`this.element.querySelectorAll('.tex')`)
- we won't load the math script on every article as we can track the script load state
- we keep the convention where `data-wikia-widget` means that a piece of content will be decorated with additional JS

Let me know if you need any help with that.

app/templates/components/widget-math.hbs
```diff
@@ -0,0 +1 @@
+<span class="tex">{{{data.tex}}}</span>
```
macbre, line 1: See https://github.com/Wikia/app/pull/15159/commits/62ba8b22d1f1ae11343c209e4b1003508f7d60c9

app/components/widget-math.js
```diff
@@ -0,0 +1,79 @@
...
+			displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
+			processEscapes: true
+		},
+		"HTML-CSS": { fonts: ["TeX"] }
+	});
+
+	// cast NodeList to an array
+	var elements = [].slice.call(window.document.body.querySelectorAll('.tex'));
+
+	window.MathJax.Hub.Queue(["Typeset", MathJax.Hub, elements]);
```
rogatty, line 80: I think these two lines are not needed.

app/components/widget-math.js
```diff
@@ -0,0 +1,79 @@
...
+	</script>`);
+
+				$script('//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/latest.js?config=TeX-MML-AM_CHTML', () => {
+					// console.log('MathJax loaded');
```
rogatty: How about cleaning up the console.logs?

app/templates/components/widget-math.hbs
rogatty, line 1: This is safe to render unescaped, I assume?

macbre: Yes, I've tested it with `4 > 5` TeX syntax.

app/components/widget-math.js
macbre: Well, they are. `renderMath` method only does rendering when the MathJax library is loaded. Hence, the first renderings need to happen via this code.

rogatty: But isn't this code also run when the MatJax library is loaded? That's why it's put in `<script type="text/x-mathjax-config">` which isn't parsed by browser. I think we do the same thing twice, here and in `renderMath`

macbre: `this.renderMath` is required to re-render the math tags (when you revisit the page with tags) or when you hit the page with tags and the MathJax is already loaded (i.e. was loaded as a part of a previous article view).

rogatty: Ok, I went through it again and now I see that the first time `renderMath` is called it does nothing.

rogatty (approved):

PR #709, opened 2018-05-18
ADEN-7162 Delay ad-engine until decision about tracking

v3rth:


app/modules/ads/legacyModule.js
```diff
@@ -448,7 +448,9 @@ class Ads {
 					delayEnabled = Boolean(adsContext.opts.delayEngine);
 				}
 
-				this.adEngineRunnerModule.run(this.adConfigMobile, this.slotsQueue, 'queue.mercury', delayEnabled);
+				M.trackingQueue.push(
```
fraszczakszymon: What about above tracking? Should we delay whole require?

nandy-andy (approved):

nandy-andy (approved):

PR #717, opened 2018-05-23
ADEN-7152 Track only first pos value

v3rth:
## Links

* https://wikia-inc.atlassian.net/browse/ADEN-7152


app/modules/ads/tracking/slot-tracker.js
```diff
@@ -1,5 +1,9 @@
 const trackingRouteName = 'special/adengadinfo';
 
+function getPosParameter({pos = ''}) {
```
nandy-andy, line 3: We can unit test it! 🙌

app/modules/ads/tracking/slot-tracker.js
```diff
@@ -1,5 +1,9 @@
...
+	return (Array.isArray(pos) ? pos : pos.split(',')).shift();
```
fraszczakszymon: `shift` modifies original array. Can we get first element without changing reference?

v3rth: You are right! thanks!

app/modules/ads/tracking/slot-tracker.js
v3rth: Added :heavy_check_mark:

nandy-andy (approved):

PR #764, opened 2018-06-13
IRIS-6105 | Fallback to zh-hant for traditional chinese languages

rogatty:
## Links
* https://wikia-inc.atlassian.net/browse/IRIS-6105
* https://github.com/Wikia/discussions-front-end/pull/221

## Description
Change i18next fallback for zh-hk, zh-mo and zh-tw to zh-hant per:
- [MessagesZh_tw.php](https://github.com/Wikia/app/blob/a90fa39e4678343e66cd31d710094c9ad6c07fba/languages/messages/MessagesZh_tw.php#L25)
- [MessagesZh_hk.php](https://github.com/Wikia/app/blob/a90fa39e4678343e66cd31d710094c9ad6c07fba/languages/messages/MessagesZh_hk.php#L21)
- [MessagesZh_mo.php](https://github.com/Wikia/app/blob/a90fa39e4678343e66cd31d710094c9ad6c07fba/languages/messages/MessagesZh_mo.php#L13)

### DIFF
--- a/app/services/i18n.js
+++ b/app/services/i18n.js
@@ -15,9 +15,14 @@ export default Service.extend({
 
 		if (fastboot.get('isFastBoot')) {
 			const fs = FastBoot.require('fs');
+			// Per MediaWiki's /languages/messages/MessagesZh_*.php
+			const traditionalChineseLanguages = ['zh-hk', 'zh-mo', 'zh-tw'];
 
 			config.translationsNamespaces.forEach((namespace) => {
 				[language, language.split('-')[0], 'en'].some((lang) => {
+					if (traditionalChineseLanguages.indexOf(lang) > -1) {
+						lang = 'zh-hant';
+					}
 					const translationPath = `dist/mobile-wiki/locales/${lang}/${namespace}.json`;
 
 					try {


app/services/i18n.js
```diff
@@ -15,9 +15,14 @@ export default Service.extend({
 
 		if (fastboot.get('isFastBoot')) {
 			const fs = FastBoot.require('fs');
+			// Per MediaWiki's /languages/messages/MessagesZh_*.php
+			const traditionalChineseLanguages = ['zh-hk', 'zh-mo', 'zh-tw'];
```
rybmat, line 19: how about specifying explicitly all Chinese fallbacks in a map to improve readability of code and then L22 change to `[map[language] || language, language.split('-')[0], 'en'].some(lang) => {...}`?

or even we could specify mappings like pt -> pt-br explicitly as well

rybmat (changes requested):

rybmat (approved):

app/services/i18n.js
rogatty: There is no pt -> pt-br, it's the opposite way and I think it's easier to just use `language.split('-')[0]`. The split is also used for languages like zh-cn, zh-sg etc. which fallback to zh.

PR #771, opened 2018-06-18
ADEM-7268 MBLB waits for recirculation-prefooter

leshniak:
## Links

https://wikia-inc.atlassian.net/browse/ADEN-7134

## Description

Due to low viewability for BLB, we forced an ad-slot to inject after the recirculation-prefooter renders.

## Reviewers

@nandy-andy 
@bkoval

### DIFF
--- a/app/components/recirculation-prefooter.js
+++ b/app/components/recirculation-prefooter.js
@@ -1,4 +1,4 @@
-import { Promise } from 'rsvp';
+import { defer } from 'rsvp';
 import fetch from 'fetch';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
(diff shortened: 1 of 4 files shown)


nandy-andy (approved):

app/components/recirculation-prefooter.js
```diff
@@ -51,6 +55,9 @@ export default Component.extend(
 				},
 				intersectionThreshold: 0
 			});
+
+			this.listRendered = defer();
```
leshniak: I know that's not the prettiest solution 😛

app/components/recirculation-prefooter.js
```diff
@@ -49,8 +53,11 @@ export default Component.extend(
 					left: 0,
 					right: 0
 				},
-				intersectionThreshold: 0
+				intersectionThreshold: 0,
+				listRendered: defer()
```
leshniak, line 57: I know that's not the prettiest solution 😛

bkoval:
How about using Ember.Evented + a service to make that look a bit less untangled? https://emberigniter.com/communication-between-distant-components/ We can discuss it offline, but I think it can make the code a bit more maintainable for you :)

nandy-andy:
@bkoval we decided to go with this solution as it's more flexible and we need to fix the issue ASAP. Feel free to join us in designing a more Ember-like solution.

PR #784, opened 2018-07-02
'just' use import from node_modules

hakubo:
It almost works.
But since we still use mefine/define it fails as webpack uses window.define for defining modules...


bkoval:
Seems like we're going with browserify for now, hence I'm closing this pull request.

PR #794, opened 2018-07-10
ADEN-7355 Rerequest one bidder in FMR X slots

dmnsobczak:


app/modules/ads/module.js
```diff
@@ -54,6 +54,7 @@ class Ads {
 		context.push('delayModules', biddersDelay);
 		events.on(events.AD_SLOT_CREATED, (slot) => {
 			bidders.updateSlotTargeting(slot.getSlotName());
+			bidders.refreshBids(slot.config);
```
fraszczakszymon: It won't be needed if we decide to register `bidWon` event somewhere in `callBidders`

dmnsobczak: Yep, refactored.

PR #830, opened 2018-07-19
XW-5119 | New typography

bkoval:
@Wikia/x-wing


app/styles/typography/_typescale.scss
```diff
@@ -0,0 +1,36 @@
+// These will live in design system after we ensure everything looks good in mobile-wiki
+.wds-text-2xs {
```
hakubo: wds-font-size-xxs {}

app/styles/typography/_typescale.scss
```diff
@@ -0,0 +1,36 @@
...
+	font-size: 10px;
```
hakubo: hmm, maybe we should use rems here?
so if users zoom in, whole page zoom in?

app/styles/typography/_font-weight-scale.scss
```diff
@@ -0,0 +1,16 @@
+// These will live in design system after we ensure everything looks good in mobile-wiki
+.wds-font-light {
```
hakubo: .wds-font-weight-light {}

app/styles/typography/_typescale.scss
rybmat: mobile-wiki does not allow to zoom, do other apps allow?

app/styles/typography/_font-weight-scale.scss
```diff
@@ -0,0 +1,16 @@
...
+	font-weight: 300;
```
rybmat: we have $wds-fandom-font-weight-* variables in DS already that have the same values. Should we use them here?

app/styles/base/_typography.scss
```diff
@@ -33,41 +38,58 @@ h3,
...
-	line-height: $wds-typescale-size-plus-1;
+	font-size: 16px;
+	font-weight: $wds-fandom-font-weight-medium;
 }
 
 h5,
 h6 {
-	font-size: $wds-typescale-size-plus-1;
-	line-height: $wds-typescale-size-plus-1;
+	font-size: 14px;
```
hakubo: should that extend font size placeholder from DS?

app/templates/components/widget-discussions.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if posts}}
-	<div class="heading-wrapper">
-		<h2 class="heading">
+	<div class="heading-wrapper wds-font-size-xs">
```
hakubo: a browser would have less work to do if that was applied on a tag
possible?

app/styles/base/_typography.scss
```diff
@@ -33,41 +38,58 @@ h3,
 h4,
 h5,
 h6 {
-	margin-bottom: 1.4rem;
-}
-
-ul,
-ol {
-	margin-left: 1.1rem;
+	margin-bottom: 18px;
```
hakubo: rems or px?

hakubo (approved):

PR #833, opened 2018-07-23
XW-5067 | Premium search result page

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/XW-5067

## Reviewers

@Wikia/x-wing


app/components/wikia-ui-components/error-info.js
```diff
@@ -10,5 +10,5 @@
 import Component from '@ember/component';
 
 export default Component.extend({
-	classNames: ['error-info', 'wds-font-size-s']
+	classNames: ['error-info', 'wds-font-size-s', 'wds-line-height-l']
```
hakubo: while we're here
tagName: ''
and add div to the template with all these clases

hakubo (approved):

PR #834, opened 2018-07-23
(PLATFORM-3570) Handle language path URLs in extractEncodedTitle

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-3570

## Description

`extractEncodedTitle` didn't take the language path into consideration, so search result URLs
ended up having two article paths in the URL, i.e. `/szl/wiki/szl/wiki/Foo`, as it thought the page
title was `szl/wiki/Foo` rather than `Foo`.

## Reviewers

@Wikia/core-platform-team @rogatty


app/utils/url.js
```diff
@@ -53,7 +53,7 @@ export function addQueryParams(url, params = {}) {
 }
 
 export function extractEncodedTitle(url) {
-	return url ? url.replace(/^(https?:\/\/[^/]+)?(\/wiki)?\//, '') : '';
+	return url ? url.replace(/^(https?:\/\/[^/]+)?(\/[a-z]{2,3}(-[a-z-]{2,12})?)?(\/wiki)?\//, '') : '';
```
rogatty: I'd move this method to https://github.com/Wikia/mobile-wiki/blob/dev/app/services/wiki-urls.js and use `this.langPath` instead of `(\/[a-z]{2,3}(-[a-z-]{2,12})?)?`

app/models/wiki/file.js
```diff
@@ -24,7 +25,7 @@ export default BaseModel.extend({
 			pageProperties = {
 				articleType: 'file',
 				fileUsageList: get(data, 'nsSpecificContent.fileUsageList')
-					.map(this.prepareFileUsageItem),
+					.map((item) => this.prepareFileUsageItem),
```
rogatty: What is this change for? `item` isn't being used.

app/services/wiki-urls.js
```diff
@@ -159,6 +159,16 @@ export default Service.extend({
...
 	},
 
+	/**
+	 * Extracts the page title from a URL by stripping the host and article path.
+	 *
+	 * @param  {string} url - URL from which to extract the title
+	 * @return {string}
+	 */
+	getEncodedTitleFromURL(url) {
+		return url ? url.replace(new RegExp(`^(https?://[^/]+)?${this.langPathRegexp}?(/wiki)?/`), '') : '';
```
rogatty, line 169: ~`this.langPathRegexp` can be an empty string so this question mark will change the matcher from greedy to lazy (https://stackoverflow.com/a/21289386/1050577) on wikis without the lang path.~

tests/unit/services/wiki-urls-test.js
```diff
@@ -341,4 +341,93 @@ module('Unit | Service | wiki-urls', (hooks) => {
 			});
 		});
 	});
+
+	test('test getEncodedTitleFromURL', (assert) => {
```
rogatty, line 345: ~I'd add another test that uses `wikiUrls.set('langPath', '/zh')` before calling `getEncodedTitleFromURL`~

app/models/wiki/file.js
Grunny: So, that `prepareFileUsageItem` has the right context, otherwise `this` is undefined in `prepareFileUsageItem`: https://stackoverflow.com/a/30148997 Missing `item` in the method call was a mistake though. :)

rogatty (approved):

ludwikkazmierczak:
Tested and Accepted

PR #840, opened 2018-07-24
XW-5146 | Premiumized mobile-wiki

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/XW-5146

@Wikia/x-wing


app/styles/base/_typography.scss
```diff
@@ -1,30 +1,35 @@
...
 blockquote {
-	border-left: .25rem solid rgba(26, 94, 184, .3);
-	color: #3a3a3a;
-	font-size: 1rem;
+	font-size: 16px;
 	font-style: italic;
-	font-weight: $font-weight-normal;
-	line-height: $line-height-base;
+	font-weight: $wds-fandom-font-weight-regular;
+	line-height: 26px;
```
hakubo: shouldn't that come from DS?

rybmat: yes it should

PR #870, opened 2018-08-08
ADEN-7483 Fix integration with ad-products bidders

dmnsobczak:


app/modules/ads/legacyModule.js
```diff
@@ -122,7 +122,6 @@ class Ads {
 				'ext.wikia.adEngine.adLogicPageParams',
 				'ext.wikia.adEngine.config.mobile',
 				'ext.wikia.adEngine.context.slotsContext',
-				'ext.wikia.adEngine.lookup.a9',
```
fraszczakszymon: We can't remove that :(. Also we need to handle fvDelay on mobile without AE3

dmnsobczak: Restored. FVDelay will be implemented in separated ticket - it's a bigger change - for now A9 in FV will be disabled.

app/modules/ads/ad-context.js
```diff
@@ -5,6 +5,7 @@ export default {
 	bidders: {
 		enabled: false,
 		timeout: 2000,
+		disabledSlots: [],
```
fraszczakszymon: https://github.com/Wikia/ad-products/pull/131/files#r209212784

fraszczakszymon (approved):

app/modules/ads/ad-context.js
fraszczakszymon: It stays empty on MW+AE3, right?

dmnsobczak: Yep, it's not necessary.

PR #881, opened 2018-08-10
XW-5114 | Remove unused components

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/XW-5114

## Reviewers

@Wikia/x-wing


latata (approved):
Maybe we should decrease numbers in assets size?

PR #889, opened 2018-08-16
(PLATFORM-3554) Handle both domains based on the current request

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-3554
* https://github.com/Wikia/discussions-front-end/pull/260

## Description

We need to change the domain related config depending on the origin of the
current request. This will still make sure if a wiki is on fandom.com that
all the config should be correct.

## Reviewers

@Wikia/core-platform-team


app/instance-initializers/config.js
```diff
@@ -52,9 +62,13 @@ export function initialize(applicationInstance) {
 	if (fastboot.get('isFastBoot')) {
 		const env = FastBoot.require('process').env;
 		const wikiaEnv = env.WIKIA_ENVIRONMENT;
+		const request = fastboot.get('request');
+
+		// Support both wikia.com and fandom.com domains during the migration period
+		config.productionBaseDomain = getProductionBaseDomain(request);
```
rogatty: Overriding `config` is a bit messy and won't propagate to browser so https://github.com/Wikia/mobile-wiki/blob/ae86bd321effece5f72bb50174634c4b4827801f/app/routes/application.js#L350 will fail.

How about we pass the base domain to all methods that need it in this file, and update `routes/application.js` to take both domains into account?

Grunny: :+1: done in 1d1b3abdf5e01dab including adding the base domain regex config var that we use in the app repo for the same purpose that will handle all envs rather than just the current prod one.

app/routes/application.js
```diff
@@ -347,7 +347,7 @@ export default Route.extend(
 					 * TODO: this regex is alright for dev environment, but doesn't work well with production
```
rogatty: This comment looks outdated.

rogatty (approved):

PR #892, opened 2018-08-16
FAN-3946 Integrate Feeds & Posts on mobile-wiki

mklucsarits:
Ticket: http://wikia-inc.atlassian.net/browse/FAN-3946
Requires: https://github.com/Wikia/app/pull/15904

This change adds the wiki's feed from the Feeds and Posts app (http://github.com/wikia/fepo) into mobile-wiki articles.


vforge (approved):
LGTM, let's wait on @Wikia/x-wing

app/routes/wiki-page.js
```diff
@@ -155,6 +156,13 @@ export default Route.extend(
 								adsModule.init(model.adsContext);
 							}
 						});
+
+						// If it's an article page and the extension is enabled, load the Feeds & Posts module
+						if (model.type === 'article' && this.get('wikiVariables.enableFeedsAndPosts')) {
```
rybmat: `model.type` equals `article` only for NS_MAIN (NS:0) articles. Some wikis use also other namespaces as content ones, like NS_PROJECT, so maybe it is worth to change this check to 
`import { isContentNamespace } from '../utils/mediawiki-namespace';`
`if (isContentNamespace(model.ns, this.get('wikiVariables.contentNamespaces')) && ... )`
?

rybmat (approved):

app/routes/wiki-page.js
```diff
@@ -155,6 +156,13 @@ export default Route.extend(
...
+							feedsAndPosts.getModule().then((fpModule) => {
+								feedsAndPosts.loadFeed(fpModule);
```
rogatty: This should be done inside `scheduleOnce('afterRender', () => {` or the DOM might not be ready yet.

app/modules/feeds-and-posts.js
```diff
@@ -0,0 +1,57 @@
...
+
+	fpPromise = new Promise((resolve, reject) => {
+		const fpScript = document.createElement('script');
+		fpScript.onload = () => {
+			resolve(window.fp);
+		};
+		fpScript.onerror = () => {
+			reject();
+		};
+		fpScript.src = '/feeds-and-posts/public/dist/lib.min.js';
```
rogatty: What about versioning / purging cache?

app/modules/feeds-and-posts.js
```diff
@@ -0,0 +1,57 @@
+import { Promise } from 'rsvp';
+
+let fpPromise = null;
+
+/**
+ * Gets the Feeds & Posts ES Module, by first appending the F&P library to the page
+ *
+ * @returns {Promise} Promise of module
+ */
+function getModule() {
```
hakubo, line 20: in ember cli you can use ember auto import to import modules from npm (also dynamically)
https://github.com/ef4/ember-auto-import#dynamic-import

can we install feeds-and-posts lib in mobile-wiki and use it instead of loading it from separate service?
also in lib.min.js there is 706KB of code - I suspect not all of it is needed, maybe using auto import would help with this as well?

hakubo: does this line mean we'll load Design System twice in mobile-wiki https://github.com/Wikia/fepo/blob/release-8/app/webpack/loaders.js#L38 ?

app/modules/feeds-and-posts.js
```diff
@@ -0,0 +1,57 @@
...
+
+	return fpPromise;
+}
+
+/**
+ * Loads the feed onto an article page
+ *
+ * @param {Module} module
+ */
+function loadFeed(module) {
```
hakubo, line 49: this could be rewritten to a pattern that is closer to React way of thinking (if I get it correctly) and it could use a component wrapper to handle the logic
here is a quick example that I did
https://ember-twiddle.com/af5a81d96e4bc50d98f30c74542dbadc?openFiles=templates.components.my-component.hbs%2Ctemplates.components.my-component.hbs

hakubo: then loading of module could also happen in this component
and we could enable/disable feeds by rendering or not the compenent

app/routes/wiki-page.js
```diff
@@ -155,6 +156,13 @@ export default Route.extend(
 								adsModule.init(model.adsContext);
 							}
 						});
+
+						// If it's an article page and the extension is enabled, load the Feeds & Posts module
+						if (model.type === 'article' && this.get('wikiVariables.enableFeedsAndPosts')) {
+							feedsAndPosts.getModule().then((fpModule) => {
```
hakubo: I was told that the name of the project is feeds

hakubo: if it is enabled will this be visible on all pages?

hakubo:
I was told this project should not be treated as an experiment but just a first phase of implementation by @xkxd and @vforge 
if that's the case I'd really love us to do it in a way that is more future proof by using more component-driven approach to loading the lib, making the lib smaller (possibly not loading DS twice) and installing it as a normal dependency. Let me know if you'd like to chat more about it

app/modules/feeds-and-posts.js
hakubo: ouch forgot that we still load AdEngine2 that stops us from using ember-auto-import - so the custom logic to load the lib has to stay for now

app/modules/feeds-and-posts.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+function getModule() {
+	if (fpPromise) {
+		return fpPromise;
+	}
+
+	fpPromise = new Promise((resolve, reject) => {
+		const fpScript = document.createElement('script');
+		fpScript.onload = () => {
+			resolve(window.fp);
```
hakubo: there are internet providers that will wrap your script in their script loading mechanism - introducing here a race condition - does not happen super often, so I'd suggest doing a check here if window.fp is available and if not - reject promise

hakubo (changes requested):

app/routes/wiki-page.js
mklucsarits: The "in-development" name is Feeds and Posts, but maybe we'll rename it to something shorter later on.

Product wants it integrated on any article page that isn't the main page.

app/modules/feeds-and-posts.js
vforge: @hakubo DS import should be resolved ~today~ with https://wikia-inc.atlassian.net/browse/FAN-3902 - the lib.min.js shouldn't load DS at all.

vforge:
@hakubo The lib right now the way we're lazy-loading the content without adding React (and other libs we're using) as another dependency.

The benefit of this approach is that we're not dependent on MW release (because it will always fetch the newest fp lib version anyway).

I would like to treat it in the same way we're treating any embedded 3rd part widget, not a library - ie. embedded twitter post.

I've added a ticket to explore other making it a regular node dependency https://wikia-inc.atlassian.net/browse/FAN-3994

app/routes/wiki-page.js
mklucsarits: I'm not sure about this. We just want the feed to load on "normal" article pages. Do you have an example of the kinds of articles loaded on other NS's?

rybmat: Don't have an example right now, however I remember we had some ticket related to such articles in mobile-wiki. Maybe someone from community support could help with examples

rogatty: Example: http://slycooper.wikia.com/wiki/Walkthrough:Turf_War!
It's an article in namespace 112. Those are normal article pages but users like to group them semantically using namespaces.
I agree with @rybmat that it's better to use `isContentNamespace`.

mklucsarits: @rybmat @rogatty Thank you, I'll make this change.

PR #898, opened 2018-08-20
XW-5222 | hide image placeholder when image is loaded

hakubo:


app/templates/components/curated-content-item.hbs
```diff
@@ -1,7 +1,9 @@
 <figure class="item-figure">
 	<div class="item-image" style={{style}}>
-		{{svg 'wds-icons-image' role='img' class='wds-icon'}}
-		<img src="{{thumbUrl}}" alt="{{model.label}}">
+		{{#unless imageLoaded}}
```
rybmat, line 3: @hakubo I was thinking about unifying this with loading images in article. We could try to reuse article-media-thumbnail component. What do you think?

hakubo: sign me up for any unification
but not sure if this should block this simple fix

rybmat (approved):

PR #905, opened 2018-08-22
XF-53-arrow-parens: deleted arrow-parens rule; updated code with eslint

Bartosz-L:
deleted arrow-parens rule from .eslintrc.js and updated code accordingly to eslint hints

### DIFF
--- a/app/mixins/alert-notifications.js
+++ b/app/mixins/alert-notifications.js
@@ -29,7 +29,7 @@ export default Mixin.create({
 
 	clearNotifications() {
 		const notifications = this.alertNotifications,
-			updatedNotifications = notifications.filter((item) => item.persistent);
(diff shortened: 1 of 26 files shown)


app/mixins/alert-notifications.js
```diff
@@ -29,7 +29,7 @@ export default Mixin.create({
 
 	clearNotifications() {
 		const notifications = this.alertNotifications;
-		const updatedNotifications = notifications.filter((item) => item.persistent);
+		const updatedNotifications = notifications.filter((item => item.persistent);
```
hakubo: I think ( should also be removed

hakubo (changes requested):

hakubo (approved):

PR #921, opened 2018-08-23
XF-53-array-callback-return

Bartosz-L:
deleted array-callback-return rule from eslint;
updated code accordingly to eslint docs


app/modules/ads/legacyModule.js
```diff
@@ -308,6 +308,7 @@ class Ads {
 			document.addEventListener(listenerSetting.eventName, () => {
 				this.trackBlocking(listenerSetting.name, GASettings[listenerSetting.name], listenerSetting.value);
 			});
+			return false;
```
hakubo: I think this map could be rewritten to forEach also

hakubo (approved):

PR #947, opened 2018-08-27
IRIS-6467 change positioning of featured video

jsutterfield:
## Links

* http://wikia-inc.atlassian.net/browse/IRIS-6467

## Description
The featured video player is positioned too low on mobile. This causes a problem where, when the user tries to close an ad, they end up clicking the link behind the player taking them away from Wikia. The fix is to update the CSS to pull that player up 5px.

Here's what the player looks like now: 
screen shot 2018-08-27 at 9 45 04 am 1
And here's what the player looks like after the CSS update:
screen shot 2018-08-27 at 9 45 15 am

## Reviewers
@Wikia/iris


xkxd (approved):

latata (commented):
I'm not sure if it's the right fix. I think we don't want to cover Global Nav by Featured Video. Default GN is 55px high.

xkxd (approved):

rybmat (approved):

PR #955, opened 2018-08-27
PLATFORM-3619 | Call cookie-syncer service to sync tracking-opt-in state between domains

jmail:
## Links
* http://wikia-inc.atlassian.net/browse/PLATFORM-3619
* https://github.com/Wikia/pandora/pull/5115

## Who might be interested
@Wikia/core-platform-team


app/components/cookiesyncer-frame.js
```diff
@@ -0,0 +1,11 @@
+import {inject as service} from '@ember/service';
+import {computed} from '@ember/object';
+import config from '../config/environment';
+import Component from '@ember/component';
+
+export default Component.extend({
+	currentUser: service(),
+	tagName: '',
+	servicesDomain: computed(() => config.app.servicesExternalHost),
```
rogatty: `config.APP.servicesExternalHost`

rogatty: See https://github.com/Wikia/ember-fandom/blob/8419b1d33907624a520aa5f5f4a4bad4cf01f0f2/addon/instance-initializers/config.js#L79

app/components/cookiesyncer-frame.js
```diff
@@ -0,0 +1,11 @@
+import {inject as service} from '@ember/service';
+import {computed} from '@ember/object';
+import config from '../config/environment';
+import Component from '@ember/component';
+
+export default Component.extend({
+	currentUser: service(),
+	tagName: '',
+	servicesDomain: computed(() => config.APP.servicesExternalHost),
+	isCookieSet: (typeof $.cookie('tracking-opt-in-status') === 'undefined' ? false : true)
```
rogatty: We're migrating away from jQuery. Please use
```window.Cookies.get('tracking-opt-in-status')```

File: app/components/cookiesyncer-frame.js [hunk 3]
```diff
@@ -0,0 +1,11 @@
+import {inject as service} from '@ember/service';
+import {computed} from '@ember/object';
+import config from '../config/environment';
+import Component from '@ember/component';
+
+export default Component.extend({
+	currentUser: service(),
+	tagName: '',
+	servicesDomain: computed(() => config.APP.servicesExternalHost),
```
hakubo: since config is not dynamic here, I don't think you need a computed prop here
servicesDomain: config.APP.servicesExternalHost - should work just fine

app/components/cookiesyncer-frame.js
hakubo: isCookieSet: typeof $.cookie('tracking-opt-in-status') !== 'undefined'
should be enough

rogatty: Also, this will fail in FastBoot because there is no `$` or `window` on server. This should be a computed property. You can use this in FastBoot:
```
fastboot: service()
...
this.get('fastboot.request.cookies')
```

app/templates/components/cookiesyncer-frame.hbs
```diff
@@ -0,0 +1,3 @@
+{{#if isCookieSet}}
+	<iframe src="https://{{servicesDomain}}/cookiesyncer/frame"></iframe>
```
rogatty: `config.APP.servicesExternalHost` already contains the protocol.

app/components/cookiesyncer-frame.js
rogatty: I think we tried this before and it didn't work.

hakubo: kk, I wonder why though...

themech (approved):

rogatty:
Don't merge before we make sure the sync is done only once.

themech (approved):

rogatty:
Closed in favour of https://github.com/Wikia/mobile-wiki/pull/1044

PR #960, opened 2018-08-28
XF-53-no-else-return

Bartosz-L:
Changed unnecessary if else statements accordingly to no-else-return rule from eslint

### DIFF
--- a/app/components/article-content.js
+++ b/app/components/article-content.js
@@ -243,13 +243,17 @@ export default Component.extend(
         // as the 'figcaption' is contained in the 'figure' element which has the '.article-media-thumbnail' class.
         if (element.closest('.portable-infobox')) {
           return 'portable-infobox-link';
-        } if (element.closest('.context-link')) {
+        }
+        if (element.closest('.context-link')) {
           return 'context-link';
-        } if (element.closest('blockquote')) {
+        }
+        if (element.closest('blockquote')) {
           return 'blockquote-link';
-        } if (element.closest('figcaption')) {
+        }
+        if (element.closest('figcaption')) {
           return 'caption-link';
-        } if (element.closest('.article-media-thumbnail')) {
+        }
+        if (element.closest('.article-media-thumbnail')) {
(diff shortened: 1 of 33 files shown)


latata (approved):

app/components/article-content.js
```diff
@@ -243,13 +243,13 @@ export default Component.extend(
         // as the 'figcaption' is contained in the 'figure' element which has the '.article-media-thumbnail' class.
         if (element.closest('.portable-infobox')) {
           return 'portable-infobox-link';
-        } else if (element.closest('.context-link')) {
+        } if (element.closest('.context-link')) {
```
hakubo: if in new line

app/mixins/wiki-page-handler.js
```diff
@@ -126,21 +125,20 @@ export default Mixin.create({
...
-    } else {
-      const wikiPageData = shoebox.retrieve('wikiPage');
-      const wikiPageError = shoebox.retrieve('wikiPageError');
-
-      // There is no way to remove stuff from shoebox, so ignore it on the consecutive page views
-      if (wikiPageError && isInitialPageView) {
-        throw wikiPageError;
-      }
+    }
+    const wikiPageData = shoebox.retrieve('wikiPage');
```
hakubo: one \n before const

app/components/article-content.js
Bartosz-L: done

app/models/user.js
```diff
@@ -34,22 +34,19 @@ export default EmberObject.extend({
...
-        } else {
-          this.logger.error('Helios connection error: ', response);
-        }
-
-        return null;
       }
+      if (response.status === 401) {
+        this.logger.info('Token not authorized by Helios');
+      }
+      this.logger.error('Helios connection error: ', response);
```
hakubo: now this will be logged always
and used to be logged only when response.status !== 401

hakubo (approved):

PR #976, opened 2018-08-29
XF-53-no-prototype-builtins

Bartosz-L:
This rule prevents calling Object.prototype methods directly from an object.

### DIFF
--- a/app/components/recirculation-prefooter.js
+++ b/app/components/recirculation-prefooter.js
@@ -105,9 +105,7 @@ export default Component.extend(
       liftigniter
         .getData(config)
         .then((data) => {
-          this.set('items', data.items.filter(
-            item => Object.prototype.hasOwnProperty.call(item, 'thumbnail') && item.thumbnail,
-          )
+          this.set('items', data.items.filter(item => item.thumbnail)
             .slice(0, recircItemsCount)
             .map((item) => {
               item.thumbnail = Thumbnailer.getThumbURL(item.thumbnail, {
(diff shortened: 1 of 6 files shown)


app/components/recirculation-prefooter.js
```diff
@@ -105,7 +105,9 @@ export default Component.extend(
       liftigniter
         .getData(config)
         .then((data) => {
-          this.set('items', data.items.filter(item => item.hasOwnProperty('thumbnail') && item.thumbnail)
+          this.set('items', data.items.filter(
+            item => Object.prototype.hasOwnProperty.call(item, 'thumbnail') && item.thumbnail,
```
hakubo: IMO this can be removed
Object.prototype.hasOwnProperty.call(item, 'thumbnail')

app/modules/video-loader.js
```diff
@@ -75,7 +75,7 @@ export default class VideoLoader {
   * @returns {class}
   */
   static getPlayerClassBasedOnProvider(provider) {
-    if (playerClassMap.hasOwnProperty(provider)) {
+    if (Object.prototype.hasOwnProperty.call(playerClassMap, provider)) {
```
hakubo: playerClassMap[provider] should be enough here

fastboot-server/logger.js
```diff
@@ -87,7 +87,7 @@ function createLogger(loggerConfig) {
   * @returns {void}
   */
   Object.keys(loggerConfig).forEach((loggerType) => {
-    if (!availableTargets.hasOwnProperty(loggerType)) {
```
hakubo, line 90: availableTargets[loggerType] should be enough here

Bartosz-L: done

PR #981, opened 2018-08-29
ADEN-7603 Load prebid library from assets

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7603


fraszczakszymon:
Closing, because we can load prebid.min.js along with global-bundle.js for now

PR #986, opened 2018-08-30
ADEN-7585 Add ctp/audio in FV tracking to DW

dmnsobczak:


app/modules/ads/tracking/jwplayer-tracker.js
```diff
@@ -63,6 +63,24 @@ export default class JWPlayerTracker {
     Object.keys(trackingEventsMap).forEach((playerEvent) => {
       player.on(playerEvent, (event) => {
         let errorCode;
+        const vastParams = event.tag ? vastParser.parse(event.tag) : null;
```
fraszczakszymon: As far as I understand we need to have those flags on `init`/`load` events. So we can't base on vast request beacuse it happens to late. Can't we use [playerState](https://github.com/Wikia/app/blob/dev/extensions/wikia/ArticleVideo/scripts/featured-video.jwplayer.ads.js#L105) for that?

dmnsobczak: You right. Added some code, now status is properly set on `ready` event (`init` is too soon).

tasks/assets-size/index.js
```diff
@@ -6,7 +6,7 @@ const assetsSizes = {
   'app.css': 85,
   'design-system.svg': 32,
   'lazy.css': 63,
-  'mobile-wiki.js': 431,
+  'mobile-wiki.js': 432,
```
nandy-andy: 🙈

fraszczakszymon (approved):

PR #1006, opened 2018-09-03
XF-663 | rootElement of ember App

Bartosz-L:
Removed ember rootElement.
Now ember should append to body initially.


rybmat (approved):
Approve as long as Google PageSpeed is ok with this change ;)

PR #1009, opened 2018-09-04
Move base meta tags to index.html so they work on error pages

rogatty:
## Description

An application error (e.g. https://maelstrom.wikia.com/wiki/Main_Page) looks bad on mobile because for some reason we don't render `head.hbs` when there is an application error. This is a quick, easy fix. There is no advantage in keeping those meta tags in `head.hbs` anyway.

## Reviewers

@Wikia/x-wing


rybmat (approved):
please fix failing test

PR #1019, opened 2018-09-05
XF-654 | Removing user-avatar component from mobile-wiki

bkoval:
@Wikia/x-wing


xkxd (approved):

app/styles/component/_widget-discussion.scss
```diff
@@ -42,15 +42,11 @@
   .post-header {
     padding-top: 12px;
 
-    .avatar {
-      display: inline-block;
-      margin-right: 5px;
+    .wds-avatar {
+      @include wds-avatar(30px, $wds-fandom-color-middle-gray);
```
rybmat: @bkoval this is why asset sizes increased, this mixin is pretty big ;) You could check if there are other usages of this mixin in mobile wiki and if they are the same then use it only once in some higher scope

rybmat (approved):

hakubo (approved):

rybmat (approved):

PR #1020, opened 2018-09-05
ADEN-7634 Request ICP with MEGA

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7634


app/modules/ads/setup.js
```diff
@@ -63,6 +63,7 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
 
   context.set('options.slotRepeater', isGeoEnabled('wgAdDriverRepeatMobileIncontentCountries'));
   context.set('slots.incontent_boxad_1.adUnit', context.get('megaAdUnitId'));
+  context.set('slots.incontent_player.adUnit', context.get('megaAdUnitId'));
```
hakubo, line 66: it looks like adEngine could accept an object with N properties to the set function

hakubo (approved):

app/modules/ads/slots.js
```diff
@@ -142,6 +142,8 @@ export default {
       incontent_player: {
         adProduct: 'incontent_player',
         avoidConflictWith: '.ad-slot',
+        autoplay: true,
+        audio: false,
```
nandy-andy, line 146: Why do we need to set this properties here? So, the MEGA ad unit knows it?

nandy-andy (approved):

PR #1023, opened 2018-09-05
XF-575 | Using appName for internal calls header

bkoval:
@Wikia/x-wing


app/models/user.js
```diff
@@ -30,7 +30,7 @@ export default EmberObject.extend({
     });
 
     return fetch(`${this.runtimeConfig.heliosInternalUrl}${queryString}`, {
-      headers: { 'X-Wikia-Internal-Request': '1' },
+      headers: { 'X-Wikia-Internal-Request': this.runtimeConfig.environment.appName },
```
hakubo: hmm I knew it'll be confusing
we'll have to rethink that...
but

whatever is set in config/environment.js
you 'just' import
import config from '..'';

rybmat (approved):

xkxd (approved):

package.json
```diff
@@ -86,7 +86,7 @@
     "ember-cli-inline-content": "0.4.1",
     "ember-cli-mirage": "0.4.9",
     "ember-cli-qunit": "4.3.2",
-    "ember-cli-sass": "7.2.0",
+    "ember-cli-sass": "7.1.7",
```
hakubo: ?

package.json
```diff
@@ -106,12 +106,12 @@
     "ember-load-initializers": "1.1.0",
     "ember-resolver": "5.0.1",
     "ember-responds-to": "1.5.0",
-    "ember-router-scroll": "0.8.0",
+    "ember-router-scroll": "0.7.1",
```
hakubo: ?

hakubo (changes requested):
package.json should not be part of this change

bkoval:
Weeeeeeeird. Removed the package.json changes anyway ;)

hakubo (approved):

PR #1040, opened 2018-09-11
FAN-4022 Do not show recirculation if Feeds & Posts is enabled

mklucsarits:
## Links

* http://wikia-inc.atlassian.net/browse/FAN-4022

## Description

For communities where Feeds & Posts is enabled, we do not want article recirculation displayed.


app/components/recirculation-prefooter.js
```diff
@@ -39,7 +39,7 @@ export default Component.extend(
 
     listRendered: null,
     isContLangEn: equal('wikiVariables.language.content', 'en'),
-    displayLiftigniterRecirculation: and('isContLangEn', 'applicationWrapperVisible'),
+    displayLiftigniterRecirculation: and('isContLangEn', 'applicationWrapperVisible') && not('wikiVariables.enableFeedsAndPosts'),
```
latata: I think it won't work. AFAIK you can't connect computed properties with logical operators. It should be something like:
```
isFeedsAndPostsDisabled: not('wikiVariables.enableFeedsAndPosts'),
displayLiftigniterRecirculation: and('isContLangEn', 'applicationWrapperVisible', 'isFeedsAndPostsDisabled'),
```

latata (commented):
One test is failing - probably because of wrong usage of computed properties

app/components/recirculation-prefooter.js
mklucsarits: That's disappointing. It would be ideal to check that wiki variable to compute the `displayLiftigniterRecirculation` property. Ember seems pretty limited in that regard.

hakubo: That is actually not true. As in React you have to install some stuff to be productive, same thing happens in Ember world as well.
There is a great tool https://github.com/kellyselden/ember-awesome-macros we're just not using it currently.

@mklucsarits

bkoval:
@mklucsarits Are we going to proceed with this one? If not, let's close it :)

mklucsarits:
Given the change of personnel in Product and the fact I've forgotten the purpose of this, I can close it. ;)

PR #1045, opened 2018-09-13
ADEN-7666 Use ad-products from ad-engine

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7666


app/styles/module/_ads.scss
```diff
@@ -1,7 +1,7 @@
 @import 'ads-floor';
 @import 'ads-mixin';
 @import 'ads-porvata';
-@import 'node_modules/@wikia/ad-products/dist/ad-products';
+@import 'node_modules/@wikia/ad-engine/dist/ad-products.css';
```
hakubo: can you check if you don't get a warning in console while building?
I think importing css is deprecated

PR #1048, opened 2018-09-14
ADEN-7490 Track labrador and queen to DW

dmnsobczak:


app/modules/ads/setup.js
```diff
@@ -144,6 +145,13 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
     adProductsUtils.mapSamplingResults(instantGlobals.wgAdDriverLABradorDfpKeyvals),
   );
 
+  // Track Labrador values to DW
+  const labradorPropValue = adProductsUtils.getSamplingResults().join(';');
+
+  if (PageTracker.isEnabled() && labradorPropValue) {
+    PageTracker.trackProp('labrador', labradorPropValue);
```
fraszczakszymon: Same here - mixed fillings about tracking in context. What about launching it in the module.js (somewhere next to adengine init)?

dmnsobczak: Done.

fraszczakszymon (approved):

PR #1064, opened 2018-09-19
ADEN-7530 Add Cheshire Cat integration to FMR X

dmnsobczak:


app/modules/ads/module.js
```diff
@@ -61,6 +62,8 @@ class Ads {
     events.on(events.PAGE_CHANGE_EVENT, this.callBidders);
     this.callBidders();
 
+    this.configureCheshireCat(instantGlobals);
```
fraszczakszymon: s/configureCheshireCat/configureBillTheLizard ?
Rebase with dev please. Should we configure BTL before labrador tracking so it will contain model variables?

app/modules/ads/module.js
```diff
@@ -77,6 +80,55 @@ class Ads {
...
+            bidderPrices.bidder_14 || 0,
+            bidderPrices.bidder_15 || 0,
+            bidderPrices.bidder_16 || 0,
+          ].join(';'),
+        });
+
+        billTheLizard.projectsHandler.enable('cheshire_cat');
+        billTheLizard.executor.register('catlapseFMR', (model, prediction) => {
+          if (prediction === 1) {
+            const slots = document.querySelectorAll("[id^='incontent_boxad_']");
```
fraszczakszymon: What do you think about `const slots = Object.keys(context.get('slots')).filter((slotName) => slotName.indexOf('incontent_boxad_') === 0`?

app/modules/ads/tracking/slot-tracker.js
```diff
@@ -45,24 +47,10 @@ function checkOptIn() {
...
-
-    if (currentSlotPrices && currentSlotPrices[bidderName]) {
-      return `${currentSlotPrices[bidderName]}not_used`;
-    }
-
-    return '';
-  }
+  const bidderPrices = targeting.getBiddersPrices(slotName);
 
   return {
```
fraszczakszymon: What about `return Object.assign({....}, bidderPrices);`?

app/modules/ads/module.js
dmnsobczak: Renamed. I will create a follow up ticket for tracking - this ticket is already quite big.

app/modules/ads/module.js
dmnsobczak: 💎 changed

app/modules/ads/tracking/slot-tracker.js
dmnsobczak: Done.

app/modules/ads/module.js
```diff
@@ -93,6 +96,56 @@ class Ads {
...
+            bidderPrices.bidder_13 || 0,
+            bidderPrices.bidder_14 || 0,
+            bidderPrices.bidder_15 || 0,
+            bidderPrices.bidder_16 || 0,
+          ].join(';'),
+        });
+
+        billTheLizard.projectsHandler.enable('cheshirecat');
+        billTheLizard.executor.register('catlapseFMR', (model, prediction) => {
+          if (prediction === 1) {
```
fraszczakszymon: We don't need to check prediction status here. It should be configured at context level (defined in `on_1`)

app/modules/ads/module.js
```diff
@@ -93,6 +96,56 @@ class Ads {
...
+            bidderPrices.bidder_12 || 0,
+            bidderPrices.bidder_13 || 0,
+            bidderPrices.bidder_14 || 0,
+            bidderPrices.bidder_15 || 0,
+            bidderPrices.bidder_16 || 0,
+          ].join(';'),
+        });
+
+        billTheLizard.projectsHandler.enable('cheshirecat');
+        billTheLizard.executor.register('catlapseFMR', (model, prediction) => {
```
fraszczakszymon: What about `catlapseIncontentBoxad`?

app/modules/ads/module.js
dmnsobczak: Right, changed.

app/modules/ads/module.js
dmnsobczak: That's even better.

fraszczakszymon (approved):

app/modules/ads/targeting.js
```diff
@@ -208,4 +208,40 @@ export default {
...
+      if (currentSlotPrices && currentSlotPrices[bidderName]) {
+        return `${currentSlotPrices[bidderName]}not_used`;
+      }
+
+      return '';
+    }
+
+    return {
+      bidder_1: transformBidderPrice('indexExchange'),
+      bidder_2: transformBidderPrice('appnexus'),
```
jacekbj, line 232: Out of curiosity, why are some of bidder numbers not used?

dmnsobczak: Those are a holes in which now removed bidders was in the past.

jacekbj (approved):

PR #1075, opened 2018-09-20
ADEN-7665 Articles in FANDOM app doesn't have ads

nandy-andy:
## Links
* https://wikia-inc.atlassian.net/browse/ADEN-7665

## Reviewers
@Wikia/iwing it seems working on my devbox. However, I have very little knowledge about Ember or Fastboot, please advice if anything can be improved.


dmnsobczak (approved):

app/services/ads.js
```diff
@@ -11,8 +11,11 @@ export default Service.extend({
   siteHeadOffset: 0,
   slotNames: null,
   noAdsQueryParam: null,
+  adsInMobileApp: null,
   noAds: computed('noAdsQueryParam', function () {
```
hakubo: computed('adsInMobileApp', 'adsInMobileApp' ...

nandy-andy: I got a comment about unfortunate name and while changing it I came up with an idea to remove `adsInMobileApp` from the service and in the application controller make a double alias! How good/bad is that? I'm pushing a code change in a minute and waiting for your advice ;)

app/services/ads.js
```diff
@@ -11,8 +11,11 @@ export default Service.extend({
   siteHeadOffset: 0,
   slotNames: null,
   noAdsQueryParam: null,
+  disableAdsInMobileApp: null,
```
nandy-andy, line 14: @hakubo says we need `computed()` here but it doesn't work throwing weird errors in console about maximum stack exceeded... any ideas? https://guides.emberjs.com/release/object-model/computed-properties/

PR #1091, opened 2018-09-27
ADEN-7701 Register Roadblock from ad-engine

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7701

### DIFF
--- a/app/modules/ads/templates/roadblock-config.js
+++ b/app/modules/ads/templates/roadblock-config.js
@@ -1,0 +1,9 @@
+export const getConfig = () => (
+  {
+    slotsToEnable: ['incontent_boxad_1', 'mobile_in_content']
+  }
+);
+
+export default {
+  getConfig,
+};
--- a/app/modules/ads/templates/roadblock-config.js
+++ b/app/modules/ads/templates/roadblock-config.js
@@ -1,8 +1,6 @@
-export default function getConfig() {
-  return {
-    slotsToEnable: [
-      'incontent_boxad_1',
-      'mobile_in_content',
-    ],
-  };
-};
+export const getConfig = () => ({
+  slotsToEnable: [
+    'incontent_boxad_1',
+    'mobile_in_content',
+  ],
+});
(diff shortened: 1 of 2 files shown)


app/modules/ads/templates/roadblock-config.js
```diff
@@ -0,0 +1,12 @@
+export const getConfig = () => (
+  {
+    slotsToEnable: [
+      'incontent_boxad_1',
+      'mobile_in_content',
+    ],
+  }
+);
+
+export default {
```
jacekbj, line 8: Out of curiosity, can it be
```
export default function getConfig() {
  return {
    slotsToEnable: [
      'incontent_boxad_1',
      'mobile_in_content',
    ],
  }
}
```?

File: app/modules/ads/templates/roadblock-config.js [hunk 2]
```diff
@@ -0,0 +1,12 @@
+export const getConfig = () => (
+  {
+    slotsToEnable: [
+      'incontent_boxad_1',
+      'mobile_in_content',
```
jacekbj: What about bottom leaderboard? How is it enabled?
I know I sometimes ask stupid/repeated questions and thank you for your patience!

app/modules/ads/templates/roadblock-config.js
fraszczakszymon: You're right 👍

app/modules/ads/templates/roadblock-config.js
fraszczakszymon: There should be only +1 slot in roadblock (along with TLB). Nef decided to enable incontent one (when AE3 is enabled)

jacekbj (approved):

app/modules/ads/templates/roadblock-config.js
fraszczakszymon: Nope, I changed it a little bit.

PR #1104, opened 2018-10-02
SEO-669 | Show banner notification about fandom.com migration

rogatty:
## Links

* https://wikia-inc.atlassian.net/browse/SEO-669
* https://github.com/Wikia/app/pull/16126
* https://github.com/Wikia/design-system/pull/276

## Description

When WikiVariables contain `fandomComMigrationNotificationBefore` or `fandomComMigrationNotificationAfter` (one at a time) we will show a banner notification with its value. When user closes the notification, we will mark it in the LocalStorage and not show the same banner again.

## Reviewers


package.json
```diff
@@ -41,7 +41,7 @@
     "compression": "1.7.3",
     "cors": "2.8.4",
     "deep-extend": "0.6.0",
-    "design-system": "github:Wikia/design-system#15.1.1",
+    "design-system": "github:Wikia/design-system#579db01fe9d8a3e69c5b414d68ec71549e4ddc3a",
```
rogatty: Remember to update after https://github.com/Wikia/design-system/pull/276 is released.

latata (approved):

xkxd (approved):

app/components/wikia-in-your-lang.js
```diff
@@ -1,10 +1,10 @@
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { getOwner } from '@ember/application';
+import localStorageConnector from '@wikia/ember-fandom/utils/local-storage-connector';
 import AlertNotificationsMixin from '../mixins/alert-notifications';
 import LanguagesMixin from '../mixins/languages';
 import WikiaInYourLangModel from '../models/wikia-in-your-lang';
-import localStorageConnector from '../utils/local-storage-connector';
```
rybmat, line 7: can we remove this util or is it used in other places?

app/services/fandom-com-migration.js
```diff
@@ -0,0 +1,68 @@
+import Service, { inject as service } from '@ember/service';
```
rybmat, line 1: how about putting it in ember-fandom? looks like something common for mobile-wiki, discussions-front-end and announcements-front-end

rogatty: We'll show it only in mobile-wiki and oasis.

app/components/wikia-in-your-lang.js
rogatty: There is no `../utils/local-storage-connector`. It worked because design-system has it in its `app/utils` which is merged with the app tree during the build. It's messy, that's why I used the explicit one.

hakubo (approved):

PR #1108, opened 2018-10-03
ADEN-7006 Implement UAP for AE3

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7006


app/modules/ads/ad-context.js
```diff
@@ -411,6 +411,9 @@ export default {
       id: 'JTKzTN3f',
     },
   },
+  slotGroups: {
+    VIDEO: ['ABCD', 'FEATURED', 'UAP_BFAA', 'UAP_BFAB', 'VIDEO'],
```
nandy-andy, line 415: Why do we need this array? :) For MEGA?

fraszczakszymon: Yes

nandy-andy (approved):

PR #1119, opened 2018-10-09
ADEN-7784 Manually call Cheshire Cat if bidders give no bids on first…

dmnsobczak:


app/modules/ads/bidders-delay.js
```diff
@@ -36,6 +36,8 @@ export default {
         resolvePromise();
         resolvePromise = null;
       }
+
+      this.biddersCalled = true;
```
fraszczakszymon: This `this` is confusing. It operates on `Ads` module like on itself

app/modules/ads/module.js
```diff
@@ -52,9 +52,11 @@ class Ads {
     const { bidders } = window.Wikia.adBidders;
     const { krux } = window.Wikia.adServices;
 
+    this.biddersCalled = false;
+
     biddersDelay.resetPromise();
     bidders.requestBids({
-      responseListener: biddersDelay.markAsReady,
+      responseListener: biddersDelay.markAsReady.bind(this),
```
fraszczakszymon: To avoid this confusion -- what about:
```
responseListener: () => {
  biddersDelay.markAsReady();
  this.biddersCalled = true;
}
```

app/modules/ads/module.js
```diff
@@ -101,62 +103,76 @@ class Ads {
   }
 
   configureBillTheLizard(instantGlobals) {
```
fraszczakszymon: Imo BTL deserves for separate file/module. It's getting bigger.

app/modules/ads/module.js
```diff
@@ -101,62 +103,76 @@ class Ads {
...
-    const { context, slotService } = window.Wikia.adEngine;
-    const { billTheLizard } = window.Wikia.adServices;
+    const { context } = window.Wikia.adEngine;
+
+    if (context.get('bidders.prebid.bidsRefreshing.enabled')) {
+      this.cheshirecatCalled = false;
+      this.cheshirecatConfig = instantGlobals.wgAdDriverBillTheLizardConfig || {};
+      this.cheshirecatPredictions = {};
+
+      context.set('bidders.prebid.bidsRefreshing.bidsBackHandler', this.callCheshireCat.bind(this));
```
fraszczakszymon: Can we make it other way around? As far as I understand we need to call cat after each incontent IF there are some changes in bids. Which means that we can set flag `shouldCallCat` on each `bidsBackHandler` and then call cat after each incontent render ended (assuming that flag has changed). In this way we won't have doubled way to call cat. What do you think?

app/modules/ads/module.js
```diff
@@ -101,62 +103,76 @@ class Ads {
...
+          if (adSlot.config.slotName === 'incontent_boxad_1' && this.biddersCalled && !this.cheshirecatCalled) {
+            this.callCheshireCat();
+          }
+        },
+      });
+    }
+  }
 
+  callCheshireCat() {
     function getNextIncontentId(predictions) {
```
fraszczakszymon: Can we declare this method outside this scope?

app/modules/ads/module.js
```diff
@@ -101,62 +103,76 @@ class Ads {
...
-            bidderPrices.bidder_14 || 0,
-            bidderPrices.bidder_15 || 0,
-            bidderPrices.bidder_16 || 0,
-          ].join(';'),
-        });
+    const bidderPrices = targeting.getBiddersPrices('mobile_in_content');
+
+    context.set('services.billTheLizard.projects', this.cheshirecatConfig.projects);
+    context.set('services.billTheLizard.timeout', this.cheshirecatConfig.timeout || 0);
+    context.set('services.billTheLizard.parameters.cheshirecat', {
```
fraszczakszymon: If we decide to separate module -- what do you think about separate method for that? i.e. `serializeBids(slotName)`

app/modules/ads/module.js
```diff
@@ -101,62 +103,76 @@ class Ads {
...
 
-          if (slot) {
-            slotService.disable(slot, 'catlapsed');
-          }
-        });
+    billTheLizard.projectsHandler.enable('cheshirecat');
+    billTheLizard.executor.register('catlapseIncontentBoxad', () => {
+      const slot = getNextIncontentId(this.cheshirecatPredictions);
+
+      if (slot) {
```
fraszczakszymon: Is it possible to be false?

app/modules/ads/bidders-delay.js
dmnsobczak: Yeah, that's true... \\/\\/\\/

app/modules/ads/module.js
dmnsobczak: ...and I don't know why I didn't thought about that :D Changed.

app/modules/ads/module.js
dmnsobczak: Discussed on Slack. There is a risk that in this approach bidders won't finish bidding before renderEnded event will be dispatched.

app/modules/ads/module.js
dmnsobczak: Yes, done.

app/modules/ads/module.js
dmnsobczak: You never know with those sneaky string variables!!... but yes, it's not possible. Removed :x

app/modules/ads/module.js
dmnsobczak: You right. Refactored to separate module.

app/modules/ads/module.js
dmnsobczak: Yep, done.

app/modules/ads/bill-the-lizard.js
```diff
@@ -0,0 +1,91 @@
...
+let cheshirecatCalled = false;
+let biddersCalled = false;
+
+export default {
+  configureBillTheLizard(instantGlobals) {
+    const { context, slotService } = window.Wikia.adEngine;
+    const { billTheLizard } = window.Wikia.adServices;
+
+    if (context.get('bidders.prebid.bidsRefreshing.enabled')) {
+      cheshirecatConfig = instantGlobals.wgAdDriverBillTheLizardConfig || {};
```
fraszczakszymon: `s/cheshirecatConfig/config` ?

app/modules/ads/bill-the-lizard.js
```diff
@@ -0,0 +1,91 @@
...
+
+      context.set('services.billTheLizard.projects', cheshirecatConfig.projects);
+      context.set('services.billTheLizard.timeout', cheshirecatConfig.timeout || 0);
+
+      billTheLizard.projectsHandler.enable('cheshirecat');
```
fraszczakszymon, line 47: What about enabling it only when `bidders.prebid.bidsRefreshing.enabled` and repeatable boxads are enabled?

app/modules/ads/bill-the-lizard.js
```diff
@@ -0,0 +1,91 @@
...
+      .then((predictions) => {
+        const identifier = this.getNextIncontentId(cheshirecatPredictions);
+        const prediction = Object.keys(predictions).map(key => `${key}=${predictions[key]}`).join(';');
+
+        cheshirecatPredictions[identifier] = prediction;
+        context.set(`services.billTheLizard.parameters.cheshirecatSlotResponses.${identifier}`, prediction);
+      });
+  },
+
+  getNextIncontentId(predictions) {
```
fraszczakszymon: I'd define it outside export, it's not needed in billTheLizard interface

app/modules/ads/bill-the-lizard.js
```diff
@@ -0,0 +1,91 @@
...
+    return `incontent_boxad_${Object.keys(predictions).length + 2}`;
+  },
+
+  serializeBids(slotName) {
```
fraszczakszymon: Same here

app/modules/ads/bill-the-lizard.js
dmnsobczak: It is done already :)
https://github.com/Wikia/mobile-wiki/blob/c7b8d79cbbc517690a594c272e2c06d6e714f063/app/modules/ads/setup.js#L133

app/modules/ads/bill-the-lizard.js
dmnsobczak: Ok, done!

app/modules/ads/bill-the-lizard.js
dmnsobczak: Done.

app/modules/ads/bill-the-lizard.js
dmnsobczak: Done.

fraszczakszymon (approved):

PR #1133, opened 2018-10-11
ADEN-7595 Autoplay info in events

pnather:
https://wikia-inc.atlassian.net/browse/ADEN-7595


app/modules/ads/tracking/player-tracker.js
```diff
@@ -35,6 +35,7 @@ function prepareData(data, playerName, eventName, errorCode = 0) {
     ctp: data.withCtp ? 1 : 0,
     audio: data.withAudio ? 1 : 0,
     video_id: data.videoId || '',
+    user_block_autoplay: !slot.config.autoplay,
```
fraszczakszymon: We should send 1/0 instead of true/false.

Imo this solution doesn't cover the problem from ticket. We need to know whether ctp was triggered by user (by setting the cookie). Queen_of_hearts/ab test impacts `slot.config.autoplay` so we can't rely on this flag to address this issue.

Loook for module called `wikia.articleVideo.featuredVideo.autoplay` -- we should take autoplay flag from `featuredVideoCookieService`

jacekbj (approved):

app/modules/ads/tracking/player-tracker.js
```diff
@@ -35,7 +35,16 @@ function prepareData(data, playerName, eventName, errorCode = 0) {
     ctp: data.withCtp ? 1 : 0,
     audio: data.withAudio ? 1 : 0,
     video_id: data.videoId || '',
+    user_block_autoplay: window.Cookies.get('featuredVideoAutoplay') === '0' ? 1 : 0,
```
fraszczakszymon: Seems not needed

app/modules/ads/tracking/player-tracker.js
```diff
@@ -36,6 +36,11 @@ function prepareData(data, playerName, eventName, errorCode = 0) {
     audio: data.withAudio ? 1 : 0,
     video_id: data.videoId || '',
   };
+
+  if ([0, 1].indexOf(data.userBlockAutoplay) > -1) {
```
fraszczakszymon: What about
```suggestion
  if ([-1, 0, 1].indexOf(data.userBlockAutoplay) > -1) {
```

fraszczakszymon (approved):

jacekbj (approved):

PR #1143, opened 2018-10-15
(PLATFORM-3719,PLATFORM-3736) Use internal proxy for calls to MediaWiki

Grunny:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-3719
* http://wikia-inc.atlassian.net/browse/PLATFORM-3736

## Description

Use internal proxy for most requests to wikis from mobile-wiki. Right now,
the majority of API calls, such as the getMobileWikiVariables call, are
going through Fastly rather than through our internal proxy. This also means
that mobile-wiki doesn't work over HTTPS on sandboxes as it was not being
served our internal certifcates for those environments.

This updates most calls to MediaWiki and makes use of our internal cache
and makes use of the shared fetch methods from ember-fandom whenever
possible.

## Reviewers


app/services/fetch.js
```diff
@@ -13,4 +13,16 @@ export default fetch.extend({
 
     this._super(...arguments);
   },
+
+  fetchFromMediaWikiAuthenticated(requestUrl, accessToken, errorClass) {
```
rogatty, line 18: This could be put in ember-fandom but I'm fine with giving I-Wing the chance to do it :wink:

rogatty (approved):
Nice shoveling :+1:

app/models/article-preview.js
```diff
@@ -1,10 +1,11 @@
 import { inject as service } from '@ember/service';
 import EmberObject from '@ember/object';
-import fetch from '../utils/mediawiki-fetch';
+import fetch from 'fetch';
 
 export default EmberObject.extend({
   wikiVariables: service(),
   wikiUrls: service(),
+  fetch: service(),
```
hakubo: since you use imported fetch
I don't think this service is needed here

app/models/article-preview.js
```diff
@@ -28,17 +29,18 @@ export default EmberObject.extend({
     });
     const FormDataClass = FastBoot.require('form-data');
     const formData = new FormDataClass();
+    const options = this.fetch.getOptionsForInternalCache(url);
```
hakubo: kinda confusing that some options are taken from service
but actually fetch is the one from import

not sure what to do though :)

app/services/fetch.js
hakubo: can you at least add a comment?

app/utils/wiki-handlers/article.js
```diff
@@ -1,4 +1,4 @@
-import fetch from '../mediawiki-fetch';
+import fetch from 'fetch';
```
hakubo, line 1: that means the LyricFind api will go through fastly
yes? is that ok?

hakubo (approved):

app/models/article-preview.js
Grunny: Yeah, I suppose we could alias the import, not sure what a good name for it would be though. :P

app/utils/wiki-handlers/article.js
Grunny: Yeah, it already was going through Fastly so nothing changes on production where this call matters, other than it might fail on sandboxes, but that's OK from our perspective for this one case. `mediawiki-fetch` was only going through the proxy if the `internalCache` option was set, which only the user model was doing, so everything else was going through Fastly. We also do this tracking call client side on Desktop, so going through Fastly for the lyrics tracking call should be OK, and it seemed like a bit more refactoring was required to use the service here, so I thought it was best to leave it as is for now.

app/services/fetch.js
Grunny: Added a TODO in 633bcb17b5ca85bfcc67173eb4194d663df23a86.

app/models/article-preview.js
Grunny: Aliased the service property when it's used with the import in the same file: 8b1013e54294e0ed61a9e323c52bacfe4cd37d13

PR #1145, opened 2018-10-15
PLATFORM-3734: passive autologin

harnash:
## Links

* https://wikia-inc.atlassian.net/browse/PLATFORM-3734
* https://github.com/Wikia/ember-fandom/pull/15

## Description

Implement passive autologin mechanisms for Safari browsers.

## Reviewers

@Wikia/core-platform-team


app/initializers/user-agent.js
```diff
@@ -0,0 +1,15 @@
+import { initialize } from 'ember-useragent/initializers/user-agent';
```
rogatty: We don't need it.

app/components/autologin-frame.js
```diff
@@ -16,6 +16,6 @@ export default Component.extend({
       return false;
     }
 
-    return typeof window.Cookies.get('autologin_done') === 'undefined' && this.currentUser.isAuthenticated;
+    return window.Cookies.get('autologin_done') === '1' && this.currentUser.isAuthenticated;
```
rogatty: I think this should be `window.Cookies.get('autologin_done') !== '1' && window.Cookies.get('autologin_done') !== '2'`
- If it's `'1'` then we already actively synced the cookies (or failed because of 3rd party cookies blocked, but there is no reason to try again)
- If it's `'2'` then we already passively synced the cookies which means there is access_token set on the other domain

app/components/autologin-passive-frame.js
```diff
@@ -0,0 +1,32 @@
...
+    if (this.fastboot.isFastBoot) {
+      // Don't create iframe in fastboot to avoid duplicate service call
+      return false;
+    }
+
+    return system === 'ios' && window.Cookies.get('autologin_done') !== '2' && !this.currentUser.isAuthenticated;
+  }),
+
+  didRender() {
+    window.addEventListener('message', this.get('reloadWindow').bind(this), false);
```
rogatty: `this.reloadWindow.bind(this)` is enough, see https://github.com/emberjs/rfcs/blob/master/text/0281-es5-getters.md

rogatty (changes requested):

app/components/autologin-passive-frame.js
harnash: fix in 1b175dedb

app/components/autologin-frame.js
harnash: That was stupid change... fix in eafb74498

rogatty (approved):

PR #1152, opened 2018-10-16
ADEN-7825 Include H3 tags in FMR X anchors for Cheshire Cat geos

dmnsobczak:


app/modules/ads/setup.js
```diff
@@ -147,6 +147,14 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
     context.set('custom.isCMPEnabled', true);
   }
 
+  const btlConfig = instantGlobals.wgAdDriverBillTheLizardConfig;
+
+  if (context.get('options.slotRepeater') && btlConfig && btlConfig.projects && btlConfig.projects.cheshirecat
```
nandy-andy: Consider extracting the conditions to a separate function, so it's readable. I wonder if it's possible that a model won't have countries list.

app/modules/ads/setup.js
```diff
@@ -147,6 +147,14 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
...
+    && btlConfig.projects.cheshirecat.some(model => utils.isProperGeo(model.countries))) {
+    context.set('slots.incontent_boxad_1.insertBeforeSelector',
+      `${context.get('slots.incontent_boxad_1.insertBeforeSelector')},.article-body h3`);
```
nandy-andy: Doesn't it mean slots will be inserted before the selector set in the context or `.article-body h3`? I think this is what we want and we can't just change the setting in context but I want to double-check if I think correctly ;)

jacekbj: We need `.article-content > h3`, otherwise it will jump into infobox ;)
https://github.com/Wikia/mobile-wiki/pull/1149/files

app/modules/ads/setup.js
dmnsobczak: Refactored. It's not possible: it's a required field.

app/modules/ads/setup.js
dmnsobczak: @jacekbj Thanks, changed!
@nandy-andy That's exactly right and that's enough. Slot repeater and slot injector will use this selector to mark places where FMR X slot can be injected and adding H3 will increase the number of possible places.

app/modules/ads/setup.js
```diff
@@ -147,6 +148,13 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
     context.set('custom.isCMPEnabled', true);
   }
 
+  const btlConfig = instantGlobals.wgAdDriverBillTheLizardConfig;
+  const insertBeforePath = 'slots.incontent_boxad_1.insertBeforeSelector';
+
+  if (context.get('options.slotRepeater') && billTheLizard.hasAvailableModels(btlConfig || {})) {
```
jacekbj: Isn't the call to `billTheLizard.hasAvailableModels` missing project name?

dmnsobczak: It is! Sorry, my bad :< Fixed.

jacekbj (approved):

dmnsobczak:
LGTM 👍

PR #1181, opened 2018-10-29
SEO-670 | SEO friendly category pages

rogatty:
## Links

* http://wikia-inc.atlassian.net/browse/SEO-670
* https://github.com/Wikia/app/pull/16144

## Description

Update category pages to be compatible with the new API. Add alphabet shortcuts and new pagination. Update styles.

## Reviewers


app/components/article-content.js
```diff
@@ -521,5 +522,23 @@ export default Component.extend(
...
+     * When users create links to category pages with query params,
+     * we block them from being crawlable. Instead of href, we use a data attribute
+     * with encoded URL which needs to be handled in JS. It reloads the page
+     * because routes/application.js don't know how to handle that.
+     * It's ok because users should get rid of old shortcuts anyway
+     * and use the built-ones instead of duplicating them.
+     */
+    handleCategoryPaginationUrls() {
+      toArray(this.element.querySelectorAll('a[data-category-url-encoded]'))
+        .forEach(anchor => anchor.addEventListener('mousedown', (event) => {
```
latata, line 536: I wonder if we could have just one event handler, just for the wrapping element.

harnash (approved):
I suggest someone else to look at it since I'm not qualified to judge this code :)

latata (approved):

app/components/article-content.js
rogatty: Those links usually exist only on category pages and only if users put them there. I think it's better to rarely have ~26 event handlers than always have an event handler that runs on **all** mousedown events and slows down **every** interaction with article content.

PR #1186, opened 2018-11-02
OPS-12119: X-Surrogate-Key header cleanup

ljagiello:
## Description
`X-Surrogate-Key` - Fastly doesn't support such header (they use `Surrogate-Key`. We want to cleanup this custom header (it's not used by our caches).

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -172,7 +172,6 @@ export default Route.extend(
           // append mobile-wiki specific key
           surrogateKey = `${surrogateKey} ${surrogateKey}-mobile-wiki`;
           fastboot.get('response.headers').set('Surrogate-Key', surrogateKey);
-          fastboot.get('response.headers').set('X-Surrogate-Key', surrogateKey);
         }
 
         // TODO remove `transition.queryParams.page`when icache supports surrogate keys


app/routes/application.js
```diff
@@ -172,7 +172,6 @@ export default Route.extend(
           // append mobile-wiki specific key
           surrogateKey = `${surrogateKey} ${surrogateKey}-mobile-wiki`;
           fastboot.get('response.headers').set('Surrogate-Key', surrogateKey);
-          fastboot.get('response.headers').set('X-Surrogate-Key', surrogateKey);
```
rogatty, line 175: @themech was there a good reason to add it in https://github.com/Wikia/mobile-wiki/pull/786/commits/74e2b583fc54dbb52e7f8357e75db91eede2882f?

themech: It was to make it compatible with other apps, that were using both `Surrogate-Key` and `X-Surrogate-Key` headers. The thing is that `Surrogate-Key` is not returned from Fastly with the response, so we thought it the `X-` version would help us with debugging/purging - you can just fetch any url and check the `X-Surrogate-Key` header to see how to purge a given wiki/app.
But luckily we can send the `Fastly-Debug` to Fastly so it returns the `Surrogate-Key`. This renders the `X-Surrogate-Key` header obsolete.

themech (approved):

PR #1203, opened 2018-11-19
IW-1283 | Delay smartbanner until we know UAP will not show up

bkoval:
@Wikia/iwing @Wikia/adeng 

The change:

We want to wait until we know for sure that UAP will not show up before showing SmartBanner to avoid a situation where SmartBanner shows up and then it's being hidden for UAP to show (content jumping).

### DIFF
--- a/app/services/smart-banner.js
+++ b/app/services/smart-banner.js
@@ -1,13 +1,13 @@
-import {computed} from '@ember/object';
+import { computed } from '@ember/object';
 import {
   and,
   equal,
   readOnly,
   or,
 } from '@ember/object/computed';
-import Service, {inject as service} from '@ember/service';
-import {track} from '../utils/track';
-import {system} from '../utils/browser';
+import Service, { inject as service } from '@ember/service';
+import { track } from '../utils/track';
+import { system } from '../utils/browser';
 import getAdsModule from '../modules/ads';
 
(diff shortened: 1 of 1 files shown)


app/services/smart-banner.js
```diff
@@ -13,6 +14,7 @@ export default Service.extend({
   dayInMiliseconds: 86400000,
   fandomAppCookieName: 'fandom-sb-closed',
   customCookieName: 'custom-sb-closed',
+  willUapNotAppear: false,
```
rybmat: can we use `willUapAppear: true`? It would be less confusing

rybmat (dismissed):

jacekbj (approved):

dmnsobczak (approved):

app/services/smart-banner.js
```diff
@@ -30,8 +33,18 @@ export default Service.extend({
     'smartBannerAdConfiguration.text',
     'isInCustomSmartBannerCountry',
     'isSystemTargetedByCustomSmartBanner',
+    'willUapNotAppear',
   ),
 
+  init() {
```
dmnsobczak, line 50: `39:3  error  Call this._super(...arguments) in init hook  ember/require-super-in-init`

latata (approved):

PR #1205, opened 2018-11-19
ADEN-7933 Adding stickyAd events

dmnsobczak:


app/modules/ads/tracking/slot-tracker.js
```diff
@@ -6,10 +6,12 @@ const onRenderEndedStatusToTrack = [
   'success',
 ];
 const onChangeStatusToTrack = [
+  'success',
```
fraszczakszymon: It will cause multiple events in DW for single slot

dmnsobczak: Ack, reverting.

fraszczakszymon (approved):

app/modules/ads/tracking/slot-tracker.js
```diff
@@ -10,8 +10,6 @@ const onChangeStatusToTrack = [
   'catlapsed',
   'error',
   'viewport-conflict',
```
jacekbj, line 12: `'viewport-conflict'` has also been removed from status.
https://github.com/Wikia/ad-engine/pull/221/files#diff-3755f3ded23b967a2b9a6ae83dee2bb8L124

jacekbj (changes requested):

app/modules/ads/tracking/slot-tracker.js
jacekbj: I'm not really sure if it needs a change.

jacekbj (approved):

app/modules/ads/tracking/slot-tracker.js
dmnsobczak: Slot can be disabled with `viewport-conflict` status, which is equal to `collapse`:
https://github.com/Wikia/ad-engine/blob/dev/src/ad-engine/services/btf-blocker-service.js#L73

PR #1212, opened 2018-11-22
ADEN-7895 Implement MOAT Yield Inteligence service

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-7895

### DIFF
--- a/app/modules/ads/module.js
+++ b/app/modules/ads/module.js
@@ -52,7 +52,7 @@ class Ads {
 
   callExternals() {
     const { bidders } = window.Wikia.adBidders;
-    const { geoEdge, krux } = window.Wikia.adServices;
+    const { geoEdge, krux, moatYi } = window.Wikia.adServices;
 
     biddersDelay.resetPromise();
     bidders.requestBids({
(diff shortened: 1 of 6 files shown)


app/modules/ads/module.js
```diff
@@ -61,6 +61,7 @@ class Ads {
 
     geoEdge.call();
     krux.call();
+    moatYi.call();
```
jacekbj, line 64: I like the pattern where the service decides whether is should run or not (instead of each app).

app/templates/head.hbs
```diff
@@ -47,6 +47,9 @@
   {{! kargo}}
   <link rel="preconnect" href="//krk.kargo.com">
   <link rel="dns-prefetch" href="//krk.kargo.com">
+  {{! moat}}
+  <link rel="preconnect" href="//mb.moatads.com" crossorigin>
```
jacekbj, line 51: Why are we doing this only on mobile-wiki?

jacekbj (approved):

PR #1272, opened 2018-12-17
 IW-1384 | mobile CSS for infobox

xkxd:
## Links
* http://wikia-inc.atlassian.net/browse/IW-1384

## Reviewers
@Wikia/iwing

### DIFF
--- a/app/components/article-content.js
+++ b/app/components/article-content.js
@@ -466,7 +466,8 @@ export default Component.extend(
 
     tabScrollHandler(scrollWrapper, scrollEl) {
       const compensation = 20;
-      const didScrollToRight = scrollEl.scrollWidth - scrollEl.scrollLeft - compensation <= scrollEl.clientWidth;
+      const computedScroll = scrollEl.scrollWidth - scrollEl.scrollLeft - compensation;
+      const didScrollToRight = computedScroll <= scrollEl.clientWidth;
       const didScrollToLeft = scrollEl.scrollLeft < compensation;
 
       if (didScrollToLeft) {
--- a/app/components/article-content.js
+++ b/app/components/article-content.js
@@ -61,6 +61,7 @@ export default Component.extend(
 
           this.handleInfoboxes();
           this.replaceInfoboxesWithInfoboxComponents();
+          this.handleScrollableTabs();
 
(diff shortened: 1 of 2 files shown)


app/components/article-content.js
```diff
@@ -453,6 +454,34 @@ export default Component.extend(
       }
     },
 
+    handleScrollableTabs() {
```
hakubo: can this be achieved without JS?

Ideally there should be no code that have to run and loop over DOM elements as this does degrade performance of loading an article.

If I get what you're trying to do here - this should be helpful http://lea.verou.me/2012/04/background-attachment-local/

xkxd: I've tried this (+ at least 3 other) CSS only approaches with backgrounds and masking and all of them had either positioning or design/theming issues that were impossible to solve from my perspective.

That's why I've ended with JS backed one. I can adjust that code to make it more performant if you have any suggestions

hakubo: Could the design be adjusted so we don't have to write code for it?
if CSS can give us 90% of what we need - I'd suggest sticking with it to avoid perf issues.

hakubo: the much simpler solution could be to show the gradient on the right
that is ~20px wide
and have a margin on tabs on the right that is also 20px wide
so when you scroll to the end - it will become invisible.

You could achieve something like this on the left side as well.

What approaches did you try? How did it look?

xkxd: I've tried
- background images/gradients with pseudo/additional elements - does not work either because of right position: absolute/scroll issues that caused the right element to scroll as well
- mask-images - because it's impossible to adjust the size and make it "repeat once, then do nothing"
- that background-attachment - because it didn't change on scroll as expected (same issues I have on this pen https://codepen.io/bennyzhao/pen/iGoBm - bottom shadow is always visible)
- some additional "fade out" solutions with opacity/gradients

All of this was also supposed to work when there's no need to scroll/shadows, so in that case, I had to cover/hide the shadows somehow.

I'll take one more approach on mobile, but tbh at this moment, I'm feeling this "shadow scroll" overcame the actual feature, because I've already spent 2 days working only on this.

PR #1274, opened 2018-12-18
Use custom workspace on QA executors and cleanup

ludwikkazmierczak:
## Links

* more relevant links

## Description

## Reviewers

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -62,8 +62,10 @@ pipeline {
         onlyStable: false,
         zoomCoverageChart: false
     }
-    cleanup{
-        deleteDir()
+    cleanup {
+        dir('node_modules') {
+          deleteDir()
+      }
     }
   }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,9 @@
 pipeline {
   agent {
-    label 'qa-executors'
+    label {
+            label 'qa-executors'
+            customWorkspace "var/lib/jenkins/workspace/mobile-wiki-pr-checks_2_EXECUTOR${env.EXECUTOR_NUMBER}"
+        }
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


mszabo-wikia (approved):
LGTM

rybmat:
ping @hakubo for awareness

Jenkinsfile
```diff
@@ -1,6 +1,9 @@
 pipeline {
   agent {
-    label 'qa-executors'
+    label {
+            label 'qa-executors'
+            customWorkspace "var/lib/jenkins/workspace/mobile-wiki-pr-checks_2_EXECUTOR${env.EXECUTOR_NUMBER}"
```
hakubo, line 5: why customWorkspace is needed?

hakubo: what if 2 PR checks are run at the same time?

Jenkinsfile
```diff
@@ -59,5 +62,8 @@ pipeline {
         onlyStable: false,
         zoomCoverageChart: false
     }
+    cleanup{
```
hakubo: missing space

Jenkinsfile
```diff
@@ -59,5 +62,8 @@ pipeline {
...
+        deleteDir()
```
hakubo: how about 
```
dir('node_modules') {
  deleteDir()
}
```
This will delete most of MB but will not require full checkout on the next run

Jenkinsfile
ludwikkazmierczak: They run in different workspace since the Executor number is on custom workspace name

Jenkinsfile
ludwikkazmierczak: 👍

Jenkinsfile
ludwikkazmierczak: this will prevent executor from keeping more workspaces then number of executor (that equals possible parallel runs). That said, it will occupy no more workspace than needed ;)

hakubo: Somehow I have a bad feeling about this :)
but what the hell, let's try!

hakubo (approved):

PR #1284, opened 2018-12-20
ADEN-8117 Fix Nielsen, Labrador and targeting setting bugs

dmnsobczak:


app/modules/ads/setup.js
```diff
@@ -76,7 +75,8 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
   context.set('services.krux.enabled', adsContext.targeting.enableKruxTargeting
     && isGeoEnabled('wgAdDriverKruxCountries') && !instantGlobals.wgSitewideDisableKrux);
   context.set('services.moatYi.enabled', isGeoEnabled('wgAdDriverMoatYieldIntelligenceCountries'));
-  context.set('services.nielsen.enabled', isGeoEnabled('wgAdDriverNielsenCountries'));
+  // ToDo: remove debug-sandbox code
+  context.set('services.nielsen.enabled', true || isGeoEnabled('wgAdDriverNielsenCountries'));
```
dmnsobczak: Make sure to remove this hardcoded `true` before deploying to production.

nandy-andy (approved):

app/modules/ads/module.js
```diff
@@ -51,25 +51,25 @@ class Ads {
   }
 
   callExternals() {
```
fraszczakszymon, line 53: Just wonder if we shoulnd't move all externals to `PAGE_RENDER_EVENT` (for example on next pv there might be different set of slots for bidders). Anyway it sounds as quite big change so let's leave it for now

app/modules/ads/setup.js
```diff
@@ -35,8 +35,7 @@ function setupAdContext(adsContext, instantGlobals, isOptedIn = false) {
     context.set('src', 'test');
   }
 
-  const labradorCountriesVariable = 'wgAdDriverLABradorTestCountries';
-  isGeoEnabled(instantGlobals[labradorCountriesVariable], labradorCountriesVariable);
+  isGeoEnabled('wgAdDriverLABradorTestCountries');
```
fraszczakszymon: Why? We need to keep it if we want to track it

PR #1314, opened 2019-01-14
IW-1507 | adjust tracking for recirc prefooter 

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/IW-1468

## Reviewers
@Wikia/iwing

### DIFF
--- a/app/templates/components/recirculation-prefooter.hbs
+++ b/app/templates/components/recirculation-prefooter.hbs
@@ -32,7 +32,7 @@
       class="recirculation-prefooter__item is-sponsored"
       style={{concat-background sponsoredItem.thumbnailUrl}}
     >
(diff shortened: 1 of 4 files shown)


app/templates/components/recirculation-prefooter.hbs
```diff
@@ -47,6 +26,26 @@
   </div>
 {{/if}}
 
+{{#if displaySponsoredContent}}
```
mszabo-wikia, line 29: is it ok for sponsored content to be below wiki articles?

rybmat: according to designs it should be in this place, see in the IW-1469 ticket

bkoval (approved):

PR #1340, opened 2019-01-24
Aden 8102

Bielik20:


dmnsobczak:
Tests are failing, please fix them :)

app/components/ads/search-top-leaderboard.js
```diff
@@ -0,0 +1,39 @@
+import Component from '@ember/component';
```
fraszczakszymon, line 1: What do you think about naming this file `search-ads` instead of `search-top-leaderboard`?

app/models/ads/ads-context.js
```diff
@@ -0,0 +1,30 @@
+import { inject as service } from '@ember/service';
```
fraszczakszymon: Same here. It seems that this file is strictly related to search page. What about `search-ads-context`?

app/templates/search.hbs
```diff
@@ -1,3 +1,4 @@
+{{ads/search-top-leaderboard}}
```
fraszczakszymon, line 1: How will it look once we have more ads on search page?

app/models/ads/ads-context.js
Bielik20: Yes, I didn't read that much into our endpoint.

bckatarzyna (approved):

PR #1361, opened 2019-02-04
ADEN-8232 Add bidder_won, bidder_won_price and scroll_y to slot tracker.

jacekbj:


app/modules/ads/tracking/slot-tracker.js
```diff
@@ -85,7 +90,10 @@ function prepareData(slot, data) {
     opt_in: checkOptIn(),
     document_visibility: utils.getDocumentVisibilityStatus(),
     // Missing:
-    // bidder_won, bidder_won_price, page_layout, rabbit, scroll_y, product_chosen
+    // page_layout, rabbit, product_chosen
+    bidder_won: slotParams.hb_bidder || '',
```
fraszczakszymon: We were already there. This approach (taking params from DOM element) is problematic when we're collapsing slot before it's created/dispatching some custom event when DOM isn't created. I believe we're storing prices received before call in `realPrices` map. Could we reuse it?

fraszczakszymon (approved):

fraszczakszymon (approved):

tests/unit/modules/ads/tracking/slot-tracker-test.js
```diff
@@ -39,6 +39,7 @@ module('Unit | Module | ads | tracking', (hooks) => {
       getSlotName: () => 'BOTTOM_LEADERBOARD',
       getStatus: () => status || 'success',
       getTargeting: () => targeting,
+      targeting,
```
fraszczakszymon, line 42: Still needed?

fraszczakszymon (approved):

dmnsobczak (approved):

PR #1385, opened 2019-02-11
IW-1311 | pause vid on 'X' click

xkxd:
## Links
* http://wikia-inc.atlassian.net/browse/IW-1311

## Description
When a user dismissed a video in `onScroll` module, it continued to play in the background, so every time next suggested video was played, the page was scrolled to top.

## Reviewers
@Wikia/iwing

### DIFF
--- a/app/modules/video-players/jwplayer.js
+++ b/app/modules/video-players/jwplayer.js
@@ -16,8 +16,9 @@ export default class JWPlayer extends BasePlayer {
 
     params.onCreate = (player) => {
       M.trackingQueue.push(() => {
+        originalOnCreate(player);
+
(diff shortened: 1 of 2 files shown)


app/modules/video-players/jwplayer.js
```diff
@@ -16,8 +16,9 @@ export default class JWPlayer extends BasePlayer {
 
     params.onCreate = (player) => {
       M.trackingQueue.push(() => {
+        originalOnCreate(player);
```
xkxd, line 19: it seems like did not call it on dev for quite a long time, so I restored it.

rybmat (approved):

PR #1400, opened 2019-02-14
ADEN-8356 Fix outstream ads

nandy-andy:
## Links
* https://wikia-inc.atlassian.net/browse/ADEN-8356


nandy-andy:
That's not right. Closing.

PR #1442, opened 2019-02-27
ADEN-8183 Set npm perm

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-8183

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -20,6 +20,9 @@ RUN addgroup -S mobile-wiki && \
     apk add --no-cache --virtual .gyp python make g++ git && \
 
 # install npm globals
+    # we are running npm install as root so we need to set unsafe-perm
+    # in order to execute all npm postinstall scripts properly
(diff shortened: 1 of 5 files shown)


xkxd (approved):

Dockerfile.dev
```diff
@@ -20,6 +20,9 @@ RUN addgroup -S mobile-wiki && \
     apk add --no-cache --virtual .gyp python make g++ git && \
 
 # install npm globals
+    # we are running npm install as root so we need to set unsafe-perm
+    # in order to execute all npm postinstall scripts properly
+    npm set unsafe-perm=true && \
```
jacekbj, line 25: We could avoid that by building dependencies as docker_user and squashing image afterwards. I don't know what are the possible consequences of this solution, probably not dangerous because it's run inside Docker context.
Let's go with it 👍

jacekbj (approved):

PR #1475, opened 2019-03-13
PLATFORM-3967: Show wikia.org migration banner

abador:
## Description
Enable messages about wikia.org migration. You need to set up flags to enable messages. Set wgWikiaOrgMigrationScheduled' or wgWikiaOrgMigrationDone to see standard message. Similar to fandom.com migration messages

### DIFF
--- a/app/mixins/wiki-page-handler.js
+++ b/app/mixins/wiki-page-handler.js
@@ -46,7 +46,6 @@ function getURL(wikiUrls, params) {
   // this is pseudo-versioning query param for collapsible sections (XW-4393)
   // should be removed after all App caches are invalidated
   query.collapsibleSections = 1;
-console.log("===============================================",query);
   return wikiUrls.build({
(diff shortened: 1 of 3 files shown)


app/mixins/wiki-page-handler.js
```diff
@@ -46,7 +46,7 @@ function getURL(wikiUrls, params) {
   // this is pseudo-versioning query param for collapsible sections (XW-4393)
   // should be removed after all App caches are invalidated
   query.collapsibleSections = 1;
-
+console.log("===============================================",query);
```
themech: looks like debugging leftovers

themech (changes requested):

app/mixins/wiki-page-handler.js
abador: missed that :/

themech (approved):

PR #1492, opened 2019-03-20
Enable surrogate keys for filepages,  PLATFORM-3973

abador:
## Links

* http://wikia-inc.atlassian.net/browse/PLATFORM-3973

## Description

Enable sending additional surrogateKeys in filepages

## Reviewers

@themech

### DIFF
--- a/app/routes/wiki-page.js
+++ b/app/routes/wiki-page.js
@@ -186,7 +186,6 @@ export default Route.extend(
 
           handler.afterModel(this, ...arguments);
         } else {
-
           if (!redirectTo) {
             redirectTo = wikiUrls.build({
               host: this.get('wikiVariables.host'),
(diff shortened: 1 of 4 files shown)


app/routes/wiki-page.js
```diff
@@ -190,6 +191,12 @@ export default Route.extend(
                 { useskin: 'oasis' },
               ),
             });
+          } else {
```
themech: I think setting surrogate key should be possible regardless of the redirect, so this should be moved one level up, before https://github.com/Wikia/mobile-wiki/pull/1492/files#diff-d77b3e7a178bee9c8affd145b2261a54R184.
Second, I think you should wrap it with the `if (fastboot.get('isFastBoot')) {` check

themech (changes requested):

app/routes/wiki-page.js
```diff
@@ -180,6 +186,7 @@ export default Route.extend(
 
           handler.afterModel(this, ...arguments);
         } else {
+
```
themech:
```suggestion
```

themech (approved):

PR #1521, opened 2019-04-04
IW-1808 | Updating DS to a version with Wikia.org logo

bkoval:
@Wikia/iwing


app/styles/_design-system.scss
```diff
@@ -10,6 +10,7 @@
 @import 'node_modules/design-system/dist/scss/wds-components/lists';
 @import 'node_modules/design-system/dist/scss/wds-components/global-footer/common';
 @import 'node_modules/design-system/dist/scss/wds-components/global-footer/breakpoint-l-down';
+@import 'node_modules/design-system/dist/scss/wds-components/global-footer-wikia-org';
```
latata, line 13: I wonder if we could load these styles conditionally only for wikia.org communities?

latata (approved):

jsutterfield (approved):

PR #1522, opened 2019-04-05
ADEN-8623 Fix ads memory leak

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-8623

### DIFF
--- a/app/modules/ads/index.js
+++ b/app/modules/ads/index.js
@@ -18,7 +18,7 @@ class Ads {
     this.engine = null;
     this.instantGlobals = null;
     this.isLoaded = false;
-    this.isFastboot = typeof window.waitForAds !== 'function';
+    this.isFastboot = typeof FastBoot === 'undefined';
     this.onReadyCallbacks = [];
   }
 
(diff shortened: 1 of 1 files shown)


app/modules/ads/index.js
```diff
@@ -387,6 +388,11 @@ class Ads {
   }
 
   waitForUapResponse(uapCallback, noUapCallback) {
```
fraszczakszymon, line 390: `waitForUapResponse` is used outside ads "namespace" (in the `smart-banner.js`). Because of that we need to handle not resolved callbacks when there is no ad-engine (e.g. in fastboot run)

app/modules/ads/index.js
```diff
@@ -18,6 +18,7 @@ class Ads {
     this.engine = null;
     this.instantGlobals = null;
     this.isLoaded = false;
+    this.isFastboot = typeof window.waitForAds !== 'function';
```
latata: ```typeof FastBoot === 'undefined'```

rybmat (approved):
I run the load test on it, looks like there is no memory leak after this change, LGTM!

latata (approved):
Just one minor comment. Also I think it would be great to find out if there are any other places where Ads constructor is called on the backend side. You can just do a simple verification using console.log in the constructor and run the app on different routes: articles/search/main page.

rybmat (approved):

PR #1533, opened 2019-04-10
IW-1915 | Temp fix for collapsible sections

latata:
## Links

* http://wikia-inc.atlassian.net/browse/IW-1915

### DIFF
--- a/app/components/article-content.js
+++ b/app/components/article-content.js
@@ -522,6 +522,11 @@ export default Component.extend(
     },
 
     toogleCollapsibleSection(header) {
+      // Temporary fix for caching issues, it can be removed after all caches expire
+      if (!header.hasAttribute('aria-controls')) {
+        return;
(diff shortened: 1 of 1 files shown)


app/components/article-content.js
```diff
@@ -522,6 +522,11 @@ export default Component.extend(
     },
 
     toogleCollapsibleSection(header) {
+      // Temporary fix for caching issues, it can be removed after all caches expire
+      if (!header.hasAttribute('aria-controls')) {
```
latata, line 526: `aria-controls` attribute is only added to headers which support collapsing. we don't want to run toggle function for headers which does not support it. 
This code make all header/sections collapsible: https://github.com/Wikia/app/pull/16937 
We just need to wait for cache invalidation. Then we can remove this temp change.

bkoval (approved):

PR #1555, opened 2019-04-25
Send internal mediawiki requests without icache proxy

rybmat:
@Wikia/iwing


mszabo-wikia (approved):

k8s/k8s-descriptor-template-verify.yaml
```diff
@@ -68,8 +68,8 @@ spec:
             value: prod
           - name: WIKIA_DATACENTER
             value: sjc
-          - name: INTERNAL_CACHE_HOST
-            value: prod.icache.service.sjc.consul
+          - name: MW_INTERNAL_HOST
+            value: mediawiki-verify
```
macbre, line 76: There is no verify on k8s

rybmat: @macbre so should we sunset verify env on mobile-wiki?

macbre: I believe so

mszabo-wikia:
This was done in https://github.com/Wikia/mobile-wiki/pull/1633

PR #1564, opened 2019-05-10
ADEN-8685 Floor Adhesion: hide after viewability

dmnsobczak:


nandy-andy (approved):

app/modules/ads/setup.js
```diff
@@ -54,6 +55,7 @@ export const adsSetup = {
     templateService.register(BigFancyAdAbove, getBfaaConfig());
     templateService.register(BigFancyAdBelow, getBfabConfig());
     templateService.register(FloorAdhesion, getOutOfPageConfig());
+    templateService.register(HideOnViewability, {});
```
Bielik20: Second argument of `templateService.register` is optional, why pass empty object?

app/modules/ads/slots.js
```diff
@@ -207,7 +207,7 @@ export const slots = {
           loc: 'footer',
           rv: 1,
         },
-        defaultTemplates: ['floorAdhesion'],
+        defaultTemplates: ['floorAdhesion', 'hideOnViewability'],
         defaultSizes: [[320, 50], [320, 100], [300, 50]],
```
Bielik20: Order.

app/modules/ads/setup.js
dmnsobczak: Yep, removed.

app/modules/ads/slots.js
dmnsobczak: That's outside the scope of the ticke... blahblah, changed :D

Bielik20 (approved):

PR #1565, opened 2019-05-14
ADEN-8686 Start Floor Adhesion after scrolling one viewport.

jacekbj:


app/modules/ads/index.js
```diff
@@ -20,6 +20,12 @@ class Ads {
     this.isLoaded = false;
     this.isFastboot = typeof FastBoot !== 'undefined';
     this.onReadyCallbacks = [];
+
+    this._readyResolve = null;
```
Bielik20: why underscore? that about `/** @private */`

Bielik20 (approved):

app/modules/ads/index.js
jacekbj: Good point. I can't wait for https://developer.mozilla.org/en-US/docs/Archive/Add-ons/Add-on_SDK/Guides/Contributor_s_Guide/Private_Properties

Bielik20 (approved):

PR #1580, opened 2019-05-24
ADEN-8779 Add page_layout tracking prop.

jacekbj:
It tracks pos_top, which is AdSlot element's distance from top.


fraszczakszymon (approved):

jacekbj:
Tests are failing. They do not stem from this ticket. They will be fixed in another PR.

PR #1603, opened 2019-05-31
ADEN-8841 Remove redundant code from out of page templates.

jacekbj:


app/styles/module/_ads.scss
```diff
@@ -145,127 +145,23 @@ $closeButtonHeight: 40px;
...
       height: $closeButtonHeight;
-      justify-content: center;
-      padding: 5px;
-      position: absolute;
-      right: 0;
       top: -($closeButtonHeight / 2);
       width: $closeButtonHeight;
     }
+
+    .advertisement-label {
```
fraszczakszymon: Is it needed? It seems it's added in the ad-engine styles

jacekbj: Good point - removed.

fraszczakszymon (approved):

PR #1604, opened 2019-06-03
ADEN-8847 Add trackConnectionDownlinkToDW

Bielik20:

### DIFF
--- a/app/modules/ads/index.js
+++ b/app/modules/ads/index.js
@@ -317,7 +317,7 @@ class Ads {
     this.trackLabradorToDW();
     this.trackDisableAdStackToDW();
     this.trackLikhoToDW();
-    this.trackConnectionDownlinkToDW();
+    this.trackConnectionToDW();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/modules/ads/index.js
```diff
@@ -357,6 +358,21 @@ class Ads {
...
+   * @private
+   */
+  trackConnectionDownlinkToDW() {
+    const { utils } = window.Wikia.adEngine;
+    const connection = navigator.connection
+      || navigator.mozConnection
+      || navigator.webkitConnection;
+
+    if (connection) {
+      pageTracker.trackProp('downlink', connection.downlink);
```
fraszczakszymon: What about naming it `bandwidth`?

jacekbj: What to you think about concatenating the info below?
- NetworkInformation.downlink (what is the distribution of network speed among users?)
- NetworkInformation.effectiveType (will make easier to create buckets for distribution of speed)
- NetworkInformation.rtt (how long it takes for a request to make a round-trip?, so I'd read it as network latency)
- NetworkInformation.saveData (how many users are using data saving?)
- NetworkInformation.type (what is the average speed of wi-fi vs cellular?)

jacekbj (approved):
Code is ✅. Please tell me what you think about extending the information.

app/modules/ads/index.js
Bielik20: I would keep the name of the actual property. Moreover because it is not the same as bandwidth: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink

fraszczakszymon: > The downlink read-only property of the NetworkInformation interface returns the effective bandwidth estimate in megabits per second

jacekbj (approved):

PR #1608, opened 2019-06-05
ADEN-8570 Change Confiant wgVariable

dmnsobczak:


app/modules/ads/setup.js
```diff
@@ -142,7 +142,7 @@ export const adsSetup = {
     context.set('options.slotRepeater', isGeoEnabled('wgAdDriverRepeatMobileIncontentCountries'));
 
     context.set('services.browsi.enabled', isGeoEnabled('wgAdDriverBrowsiCountries'));
-    context.set('services.confiant.enabled', isGeoEnabled('wgAdDriverConfiantCountries'));
+    context.set('services.confiant.enabled', isGeoEnabled('wgAdDriverConfiantMobileCountries'));
```
nandy-andy: Wouldn't it be more solid if we keep check for both variables? Then remove the old one 24h after the release on two new ones. I think if we do this like that we won't introduce any disturbance in the integration.

dmnsobczak: If you worried about the cache then it's fine. But you right, there might be a case if someone edit old variable. I'll add old variable here too, thanks! :)

jacekbj (approved):

PR #1609, opened 2019-06-06
ADEN-8768 Change Mobile-Wiki build to use Awesome TypeScript Loader

Bielik20:


app/modules/ads/index.js
```diff
@@ -42,19 +42,49 @@ class Ads {
...
+      window.Wikia.adServices = module;
+      window.Wikia.adBidders = module;
+      return module;
+    });
+  }
+
+  /**
+   * @private
+   */
+  static getShouldStartAdEngine() {
```
dmnsobczak, line 77: If you really want two verbs, how about `getLoadAdEngine`?

dmnsobczak (approved):

PR #1610, opened 2019-06-07
ADEN-8871 Setting context overscrolled status

dmnsobczak:


app/modules/ads/tracking/slot-tracker.js
```diff
@@ -10,6 +10,7 @@ export const onChangeStatusToTrack = [
   'catlapsed',
   'error',
   'hivi-collapse',
+  'overscrolled',
```
fraszczakszymon: Seems not needed. `emitEvent` on AdSlot doesn't trigger statusChange, but customEvent (which is always tracked).

dmnsobczak: 👍

fraszczakszymon (approved):

PR #1612, opened 2019-06-10
IW-2107: add license description to wiki pages

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/IW-2107
* https://github.com/Wikia/design-system/pull/352
* https://github.com/Wikia/app/pull/17151

## Description

Show license info in article footer

## Reviewers

@Wikia/iwing

### DIFF
--- a/app/templates/components/article-wrapper.hbs
+++ b/app/templates/components/article-wrapper.hbs
@@ -100,5 +100,8 @@
         {{/each}}
       {{/collapsible-menu}}
     {{/if}}
+    <div class="license-description wds-font-size-s wds-leading-tight">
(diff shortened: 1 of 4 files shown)


jsutterfield (approved):

mszabo-wikia (approved):

app/templates/components/article-wrapper.hbs
```diff
@@ -100,5 +100,8 @@
         {{/each}}
       {{/collapsible-menu}}
     {{/if}}
+    <div class="license-description wds-font-size-s wds-leading-tight">
```
latata, line 103: Don't you think it would be better to have all the styling in scss?

latata (approved):

app/templates/components/article-wrapper.hbs
rybmat: for me it does not matter, the only difference I see is that it does not make css bigger ;)

PR #1616, opened 2019-06-11
MobileWiki FastBoot fix

Bielik20:


app/modules/ads/index.js
```diff
@@ -42,8 +42,8 @@ class Ads {
       return adsPromise;
     }
 
-    if (window.isNotFastboot !== true) {
-      return new Promise((resolve, reject) => reject());
+    if (typeof FastBoot !== 'undefined') {
```
rybmat, line 45: how about extracting https://github.com/Wikia/mobile-wiki/pull/1616/files#diff-34a2814942202b24a3dd9801e609c9f3R21 to separate static method and use it here to not duplicate the same condition?

rybmat (approved):
one suggestion, but overall LGTM

PR #1645, opened 2019-06-26
IW | Update Node.js to v10 LTS

mszabo-wikia:
## Description
The previous Node.js upgrade from 8.9.1 -> 8.16.0 was unironically a Great Success™️. Let's go to the current LTS release - v10.

This PR updates the Node.js runtime and also updates CI to ensure that it runs the same version of Node.


rybmat (approved):
LGTM, but 2 notes from me:
1) Apart from selenium tests lets do some manual regression testing since it is kind of big update
2) lets wait with release till at least Monday, we will be able to see what was the impact of icache bypass release and what is the impact of this update

PR #1671, opened 2019-07-04
IW-2154 | track impressions of sponsored item in recirc

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/IW-2154
* https://github.com/Wikia/design-system/pull/360

## Description

## Reviewers

@Wikia/iwing

### DIFF
--- a/app/components/recirculation-prefooter.js
+++ b/app/components/recirculation-prefooter.js
@@ -2,7 +2,7 @@ import { defer } from 'rsvp';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { reads, and } from '@ember/object/computed';
-import { computed } from '@ember/object';
+import { computed, observer } from '@ember/object';
 import { run } from '@ember/runloop';
 import InViewportMixin from 'ember-in-viewport';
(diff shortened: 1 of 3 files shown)


app/components/recirculation-prefooter.js
```diff
@@ -50,6 +51,17 @@ export default Component.extend(
       }) : this.sponsoredItem.thumbnail;
     }),
 
+    sponsoredItemObserver: observer('sponsoredItem', 'sponsoredItemDisplayed', function () {
+      if (this.sponsoredItem && this.sponsoredItemDisplayed) {
+        // this tracking needs to be done in observer since the sponsored item is fetched async
```
mszabo-wikia, line 56: hm, maybe we should return the promise from sponsored-content service to allow clients to register a handler for it

rybmat: I thought about that, but then it would be tracked when the item is fetched, not displayed. Given the fact that it is fetched only once for both, footer and search modal, then it would be tracked twice on the impression of first of the two components

mszabo-wikia (approved):

PR #1683, opened 2019-07-11
ADEN-9024 Call ICBM on mobile

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-9024

### DIFF
--- a/app/modules/ads/ad-context.js
+++ b/app/modules/ads/ad-context.js
@@ -418,6 +418,10 @@ export const defaultAdContext = {
       enabled: false,
       propertyId: 'd-aIf3ibf0cYxCLB1HTWfBQOFEA',
     },
+    instantConfig: {
+      endpoint: 'https://services.wikia.com/icbm/api/config?app=mobile-wiki',
+      fallbackConfigKey: 'fallbackInstantConfig',
+    },
     krux: {
       enabled: false,
       id: 'JTKzTN3f',
--- a/app/modules/ads/ad-context.js
+++ b/app/modules/ads/ad-context.js
@@ -420,7 +420,6 @@ export const defaultAdContext = {
     },
     instantConfig: {
       endpoint: 'https://services.wikia.com/icbm/api/config?app=mobile-wiki',
-      fallbackConfigKey: 'fallbackInstantConfig',
     },
     krux: {
       enabled: false,
--- a/app/modules/ads/app-configuration.js
+++ b/app/modules/ads/app-configuration.js
@@ -12,11 +12,11 @@ export default class AppConfiguration {
   }
 
   get(key, defaultValue = null) {
-    if (this.config && Object.prototype.hasOwnProperty.call(this.config, key)) {
+    if (this.config && key in this.config) {
       return this.config[key];
     }
 
-    if (this.instantGlobals && Object.prototype.hasOwnProperty.call(this.instantGlobals, key)) {
+    if (this.instantGlobals && key in this.instantGlobals) {
       return this.instantGlobals[key];
     }
 
--- a/app/modules/ads/app-configuration.js
+++ b/app/modules/ads/app-configuration.js
@@ -12,11 +12,11 @@ export default class AppConfiguration {
   }
 
   get(key, defaultValue = null) {
-    if (this.config && this.config.hasOwnProperty(key)) {
+    if (this.config && Object.prototype.hasOwnProperty.call(this.config, key)) {
       return this.config[key];
     }
 
-    if (this.instantGlobals && this.instantGlobals.hasOwnProperty(key)) {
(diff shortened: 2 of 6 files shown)


app/modules/ads/ad-context.js
```diff
@@ -418,6 +418,10 @@ export const defaultAdContext = {
       enabled: false,
       propertyId: 'd-aIf3ibf0cYxCLB1HTWfBQOFEA',
     },
+    instantConfig: {
+      endpoint: 'https://services.wikia.com/icbm/api/config?app=mobile-wiki',
```
mszabo-wikia, line 422: what about fastboot?

mszabo-wikia (changes requested):

app/modules/ads/ad-context.js
fraszczakszymon: It isn't called on fastboot side. Instant config service is a part of ad-engine bundle (which is also not loaded on server side).

mszabo-wikia (approved):

app/modules/ads/app-configuration.js
```diff
@@ -0,0 +1,33 @@
...
+
+    return instantConfig.getConfig().then((config) => {
+      this.config = config;
+
+      return config;
+    });
+  }
+
+  get(key, defaultValue = null) {
+    if (this.config && Object.prototype.hasOwnProperty.call(this.config, key)) {
```
Bielik20: What about `this.config.hasOwnProperty(key)` or `key in this.config`?

app/modules/ads/app-configuration.js
```diff
@@ -0,0 +1,33 @@
...
+      return this.config[key];
+    }
+
+    if (this.instantGlobals && Object.prototype.hasOwnProperty.call(this.instantGlobals, key)) {
```
Bielik20: ditto

app/modules/ads/index.js
```diff
@@ -133,15 +132,15 @@ class Ads {
 
     const { utils } = window.Wikia.adEngine;
 
-    this.instantGlobals = instantGlobals;
+    this.triggerInitialLoadServices(mediaWikiAdsContext, instantGlobals, isOptedIn)
```
Bielik20, line 135: I wonder how it worked before your changes 😕

app/modules/ads/app-configuration.js
```diff
@@ -0,0 +1,33 @@
...
+    this.instantGlobals = instantGlobals;
+
+    return instantConfig.getConfig().then((config) => {
+      this.config = config;
+
+      return config;
+    });
+  }
+
+  get(key, defaultValue = null) {
```
Bielik20, line 14: I see you have smuggled this `defaultValue` argument. I know it is related to this additional type check, but still I have mixed feelings about it. We had similar discussion about `AdEngine.Context` (you, Jacek and me), and I believe we had agreed that `||` operator is more verbose and does not require knowledge about the definition of function. But again it creates problems with falsy values... I don't know, perhaps it is ok. I would like to here what Jacek has to say since he participated in the original discussion as well @jacekbj

app/modules/ads/app-configuration.js
```diff
@@ -0,0 +1,33 @@
+export default class AppConfiguration {
```
Bielik20, line 1: I feel like I am missing something. I thought that kind of logic will be encapsulated within the ad-engine (hidden) so that client app doesn't know if it is using instant globals or instant config.

app/modules/ads/app-configuration.js
fraszczakszymon: Yes we did and we decided to use lodash in app/ae3. However it's not available in mobile-wiki and I don't want to include lodash in the ember build. Falsy values are not ok :(. We need to have possibility to define falsy values from instant config.

app/modules/ads/index.js
fraszczakszymon: What do you mean? It wasn't async before.

app/modules/ads/app-configuration.js
fraszczakszymon: It's a wrapper for combining instant-config and instant-globals. Once we switch fully to instant-config we'll be able to remove this class.

app/modules/ads/app-configuration.js
fraszczakszymon: Linter says:
> Do not access Object.prototype method 'hasOwnProperty' from target object. (no-prototype-builtins)

🤷‍♂️

app/modules/ads/app-configuration.js
fraszczakszymon: Linter says:
> Do not access Object.prototype method 'hasOwnProperty' from target object. (no-prototype-builtins)

🤷‍♂️

Bielik20 (approved):

dmnsobczak (approved):

PR #1697, opened 2019-07-19
[ADEN-9068] GAM lazy loading enabled

Loczi94:
Explanation:
To enable GAM lazy loading:
- class 'hide' has to be removed for bottom_leaderboard, top_boxad and incontent_boxad_1,
- flag disablePushOnScroll for incontent-boxad_1 has to be true,
- option incontentBoxad1EagerLoading has to be on.

### DIFF
--- a/app/modules/ads/slots.js
+++ b/app/modules/ads/slots.js
@@ -171,7 +171,6 @@ export const slots = {
             'targeting.pos': ['incontent_boxad'],
           },
           insertBelowScrollPosition: true,
-          disablePushOnScroll: true,
         },
         slotShortcut: 'f',
         sizes: [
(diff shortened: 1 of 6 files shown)


app/modules/ads/slots.js
```diff
@@ -171,6 +171,7 @@ export const slots = {
             'targeting.pos': ['incontent_boxad'],
           },
           insertBelowScrollPosition: true,
+          disablePushOnScroll: true,
```
bckatarzyna: You have already set it in setup.js:158, so you can remove this line (as it's only for GAM lazy loading not for every incontent_boxad_1)

bckatarzyna (changes requested):

bckatarzyna (approved):

bckatarzyna (approved):

dmnsobczak (approved):

PR #1699, opened 2019-07-19
ADEN-8839 Plug in tracking Postmessage messages to DW.

jacekbj:


app/modules/ads/tracking/postmessage-tracker.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return postmessageTracker
+    .add(trackingPayloadValidationMiddleware)
+    .register(
+      (message) => {
+        const { target, payload } = message;
+
+        switch (target) {
+          case TrackingTarget.GoogleAnalytics: {
+            track(Object.assign({}, payload, {
```
Bielik20, line 22: why not spread?

app/modules/ads/tracking/postmessage-tracker.js
```diff
@@ -0,0 +1,41 @@
...
+              eventName: 'trackingevent',
+              trackingMethod: TrackingMethod.ga,
+            }), false);
+            break;
+          }
+          case TrackingTarget.DataWarehouse:
+            track(Object.assign({}, payload, {
```
Bielik20, line 29: ditto

app/modules/ads/tracking/postmessage-tracker.js
jacekbj: Spread is not supported in this place. Adding it results in an error.

Bielik20 (approved):

PR #1703, opened 2019-07-24
ADEN-8997 | Track average session scroll speed

bckatarzyna:


fraszczakszymon (changes requested):
It is going to fail when sessionStorage is not available. We have `localCache`/`localStorage` module in the `Wikia/ad-engine`, would it be useful for that case?

Also, I believe that tickets is not about browser's session, but user's session. Our own implementation of window.session_id, etc. It has different logic than sessionStorage. AFAIR it expires/refreshes after 30 mins of inactivity (non-visiting next pv).

fraszczakszymon:
I would also move scroll-tracker with session/cookie management to ad-engine

app/modules/ads/index.js
```diff
@@ -367,7 +370,7 @@ class Ads {
    * @private
    */
   triggerPageTracking() {
-    scrollTracker.initScrollSpeedTracking();
+    this.initScrollSpeedTracking();
```
fraszczakszymon, line 374: It is going to add new `touchstart` event listener on each pageview

fraszczakszymon: https://github.com/Wikia/ad-engine/pull/464/files#r308563453

fraszczakszymon: OK, sorry. I see reset method

app/utils/track.js
```diff
@@ -200,10 +200,9 @@ export function trackPageView(isInitialPageView, uaDimensions) {
 /**
  * Tracks scrollY position at given time
  * @param {number} time
+ * @param {number} scrollY
  */
-export function trackScrollY(time) {
-  const scrollY = window.scrollY || window.pageYOffset;
-
+export function trackScrollY(time, scrollY) {
```
fraszczakszymon: Where is it used now?

fraszczakszymon (approved):

fraszczakszymon (approved):

PR #1704, opened 2019-07-25
ADEN-8996 Adding Viewability Counter service

dmnsobczak:


app/modules/ads/index.js
```diff
@@ -368,12 +368,39 @@ class Ads {
...
+  trackViewabilityToDW() {
+    const { events, eventService } = window.Wikia.adEngine;
+    const { viewabilityCounter } = window.Wikia.adServices;
+
+    pageTracker.trackProp('session_viewability_all', viewabilityCounter.getViewability());
+    pageTracker.trackProp('session_viewability_tb', viewabilityCounter.getViewability('top_leaderboard'));
+    pageTracker.trackProp('session_viewability_icb', viewabilityCounter.getViewability('incontent_boxad'));
+
+    eventService.on(events.AD_SLOT_CREATED, (slot) => {
+      const slotName = slot.getSlotName().substring(0, 16) === 'incontent_boxad_'
```
fraszczakszymon: 🔴 Let's create flag in slot's context definition like `viewabilityCounterId` and then:
```suggestion
      const id = slot.getConfigProperty('viewabilityCounterId') || slot.getSlotName();
```

fraszczakszymon: Also, what do you think about creating method `init` in the `viewabilityCounter`. So it takes list of defined slot names (from context) and updates their status. Then there will be no need to handle this kind of "updating" logic on platform's side.

dmnsobczak: Flag added. I also added `init()` method, but I skipped idea of slots list (this could be problematic with ICB) and added check preventing multiple inits on SPA pages.

fraszczakszymon (approved):

PR #1726, opened 2019-08-06
ADEN-9083 Use ICBM for floor adhesion delay/timeout

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-9083


jacekbj (approved):

Bielik20 (approved):
💎

PR #1760, opened 2019-08-23
ADEN-9244 Fix incorrect behaviour of scroll tracker.

bckatarzyna:


app/modules/ads/index.js
```diff
@@ -463,9 +461,10 @@ class Ads {
    * @private
    */
   initScrollSpeedTracking() {
-    const { scrollTracker } = window.Wikia.adServices;
+    const { ScrollTracker } = window.Wikia.adServices;
 
-    scrollTracker.initScrollSpeedTracking('application-wrapper');
+    this.scrollTracker = new ScrollTracker([0, 2000, 4000], 'application-wrapper');
```
Bielik20: This could be configurable from icbm. Do you think it would be necessary?

Bielik20 (changes requested):

app/modules/ads/index.js
jacekbj: Yes, it could. I think that if a need to change it arises then we will move it to ICBM. The thing is that these values must be a multiplication of 2s (that's result of scroll speed calculator design). With the current experiment there will be no need to change.

Bielik20 (approved):

Bielik20 (approved):

PR #1762, opened 2019-08-23
SER-3453 Track addtional data to discern crosswiki search

mateuszsip:
## Links

* https://wikia-inc.atlassian.net/browse/SER-3453
* https://wikia-inc.atlassian.net/browse/DE-4494


package.json
```diff
@@ -35,7 +35,7 @@
   "dependencies": {
     "@wikia/ad-engine": "git+https://github.com/Wikia/ad-engine.git#v41.9.3",
     "@wikia/ember-fandom": "github:Wikia/ember-fandom#20adaf9b214b67c1046090b6caed91aec92c472a",
-    "@wikia/search-tracking": "github:Wikia/search-tracking#1.2.0",
```
mateuszsip, line 38: Use version tag after https://github.com/Wikia/search-tracking/pull/10 is released

app/controllers/search.js
```diff
@@ -43,15 +43,17 @@ export default Controller.extend({
       searchPhrase: this.inputPhrase,
       clicked: {
         type: 'article', // currently the only displayed type in the search
-        id: result.id,
+        id: result.wikiId + '_' + result.id,
```
slayful: Won't this break existing tracking?

app/controllers/search.js
```diff
@@ -43,15 +43,17 @@ export default Controller.extend({
...
         thumbnail: false, // we do not show thumbnails on SRP right now
-        wikiId: result.wikiId,
+      },
+      filters: {
+          searchType: this.model.getScope()
       },
       target: 'redirect',
       app: 'mw-mobile',
-      siteId: result.wikiId,
+      siteId: this.wikiVariables.id,
```
slayful, line 56: So we pass the ID of the wiki we're currently on?

app/controllers/search.js
slayful: `result` is the object from the model or from the view?

app/controllers/search.js
```diff
@@ -43,15 +43,17 @@ export default Controller.extend({
       searchPhrase: this.inputPhrase,
       clicked: {
         type: 'article', // currently the only displayed type in the search
-        id: result.id,
+        id: result.wikiId + '_' + result.id,
         title: result.title,
         position: result.position + 1, // +1 since we need to start with 1 instead of 0
         thumbnail: false, // we do not show thumbnails on SRP right now
```
slayful, line 49: I don't see anything that would indicate that it's a cross-wiki article search other than the ID including the wiki ID of possibly another wiki. If that's the only indication how do we differentiate clicks on articles from wiki the user is currently on in both search modes?

app/controllers/search.js
```diff
@@ -69,9 +71,11 @@ export default Controller.extend({
     const batchEnd = batchBegin + batchSize;
     const payload = {
       searchPhrase: this.inputPhrase,
-      filters: {}, // there is no way in mobile-wiki to set any filter
+      filters: {
+        searchType: this.model.getScope()
+      }, // there is no way in mobile-wiki to set any filter
```
slayful: Well now there is, right? :P

slayful (changes requested):
I have some concerns about differentiating between cross-wiki and on-wiki article search

app/controllers/search.js
mateuszsip: https://wikia-inc.atlassian.net/browse/DE-4494 - that's the blocker I mentioned on daily - we agreed on this contract during meeting with DE last week.

app/controllers/search.js
mateuszsip: Yes, and `clicked.id` is a composite key combining wiki id and page id.

slayful (approved):
My concerns have been addressed, the filter is included in the payload and the data will be migrated.

app/controllers/search.js
mateuszsip:
```suggestion
      },
```

rybmat (approved):

PR #1770, opened 2019-08-30
ADEN-9200 Add slot click tracking to ihi2.

jacekbj:


app/modules/ads/setup.js
```diff
@@ -201,6 +201,9 @@ export const adsSetup = {
       context.set('slots.floor_adhesion.clickPositionTracking', true);
       context.set('slots.floor_adhesion.forceSafeFrame', false);
     }
+    if (instantConfig.get('icInvisibleHighImpact2ClickPositionTracking')) {
+      context.set('slots.invisible_high_impact_2.clickPositionTracking', true);
```
fraszczakszymon: What about
```suggestion
      context.set('slots.invisible_high_impact_2.clickPositionTracking', instantConfig.get('icInvisibleHighImpact2ClickPositionTracking'));
```
?

fraszczakszymon: (without if statement)

fraszczakszymon (approved):

app/modules/ads/setup.js
jacekbj: Good one, applied.

PR #1771, opened 2019-09-02
ADEN-9287 Change setting `slots.floor_adhesion.forceSafeFrame`.

jacekbj:
icFloorAdhesionClickPositionTracking takes precedence over icFloorAdhesionForceSafeFrame.

### DIFF
--- a/app/modules/ads/setup.js
+++ b/app/modules/ads/setup.js
@@ -194,16 +194,14 @@ export const adsSetup = {
     context.set('slots.incontent_player.videoAdUnit', context.get('vast.dbNameAdUnitId'));
 
     context.set('slots.floor_adhesion.disabled', !instantConfig.isGeoEnabled('wgAdDriverMobileFloorAdhesionCountries'));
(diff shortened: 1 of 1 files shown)


app/modules/ads/setup.js
```diff
@@ -196,10 +196,17 @@ export const adsSetup = {
     context.set('slots.floor_adhesion.disabled', !instantConfig.isGeoEnabled('wgAdDriverMobileFloorAdhesionCountries'));
     if (instantConfig.get('icFloorAdhesionClickPositionTracking')) {
       context.set('slots.floor_adhesion.clickPositionTracking', true);
```
fraszczakszymon: We can move this line above L205 and get rid of this if statement.

fraszczakszymon (approved):

app/modules/ads/setup.js
jacekbj: 💎

fraszczakszymon (approved):

PR #1782, opened 2019-09-12
post-search-results component

vforge:
- Add `post-search-results` component (+subcomponent, +styles, etc)
- Component queries unified search for posts that match query
- Link to current community visible on internal search
- I18n everything
- Component will be displayed on communities with Quizzes **and on ALL non-prod envs**.

@Wikia/cake


app/components/post-search-results.js
```diff
@@ -0,0 +1,120 @@
...
+  'battlefield.fandom.com',
+  'dragonage.fandom.com',
+  'acecombat.fandom.com',
+  'borderlands.fandom.com',
+  'pixelgun-wiki.fandom.com',
+];
+
+
+export default Component.extend({
+  fetchService: service('fetch'),
```
JoshuaRogan, line 56: This is so interesting. Just how this whole system works. Never touched ember.

app/components/post-search-results-item.js
```diff
@@ -0,0 +1,21 @@
...
+    const stats = this.post.stats;
+
+    return type === 'quiz' && stats && stats.takes && parseInt(stats.takes, 10) > 0;
+  }),
+
+  imageThumbnail: computed('post.image', function () {
+    const image = this.post.image;
+
+    return window.Vignette ? window.Vignette.getThumbURL(image, {
+      mode: window.Vignette.mode.zoomCrop,
```
JoshuaRogan: Do you have to check for these? or can we safely assume those props will be there?

app/components/post-search-results.js
```diff
@@ -0,0 +1,120 @@
...
+
+    const queryParams = {
+      query,
+      page: 0,
+      lang: this.wikiVariables.language.content,
+      limit: 3,
+    };
+
+    if (this.isInternal) {
+      queryParams.wikiId = this.wikiVariables.id;
```
JoshuaRogan, line 116: What does `isInternal` mean?

config/bundlesize.js
```diff
@@ -5,15 +5,15 @@ const assetsFolder = 'mobile-wiki/assets';
 module.exports = {
   'mobile-wiki.js': {
     pattern: `${assetsFolder}/mobile-wiki-*.js`,
-    limit: '478KB',
+    limit: '485KB',
```
JoshuaRogan, line 8: I kind of like that you have to bump these when you change the. kind of...

JoshuaRogan (approved):

app/components/post-search-results.js
```diff
@@ -0,0 +1,134 @@
...
+  // 'dragonage.fandom.com',
+  // 'acecombat.fandom.com',
+  // 'borderlands.fandom.com',
+  // 'pixelgun-wiki.fandom.com',
+];
+
+
+export default Component.extend({
+  fetchService: service('fetch'),
+  logger: service(),
```
mklucsarits, line 57: Is this a type of dependency injection? `service()` knows what logger to use?

app/components/post-search-results.js
vforge: Yeah. It's a lot of :magic:

app/components/post-search-results-item.js
vforge: It's copied over from another component.

app/components/post-search-results.js
vforge: Internal query => query current wiki. Same param is used in regular search component.

app/styles/component/_post-search-results-item.scss
```diff
@@ -0,0 +1,46 @@
...
+    font-weight: $wds-font-weight-normal;
+    line-height: 1;
+
+    & > svg {
+      margin-right: 6px;
+    }
+  }
+
+  &__image {
+    border: solid 1px #bed2cf;
```
mklucsarits: I _think_ this is supposed to be `$wds-fandom-color-mid-light-gray` even though Green is off by 1 number.

app/components/post-search-results.js
vforge: Apparently. Because `logger.js` exists in `app/services`

config/bundlesize.js
vforge: Same.

app/styles/component/_post-search-results-item.scss
vforge: 👍 Good finiding. This component will go through Design Review once it's merged and enabled on on keikotest wiki on prod.

vforge: ff5cd58

mklucsarits (approved):

app/components/post-search-results.js
mszabo-wikia: As we're here, may be worth to rename it to avoid confusion.

app/components/post-search-results-item.js
rybmat: I don't think the check is needed, the vignette-js lib is loaded here https://github.com/Wikia/mobile-wiki/blob/8ddea9e2f7cca5f40bd969c23ec092b241d73c4b/ember-cli-build.js#L147, there are actual several places where it is used without check. Also, instead of using `window.Vignette` you could use 
`import Thumbnailer from '../modules/thumbnailer';`
and then `Thumbnailer.getThumbURL(...)`

app/components/post-search-results.js
```diff
@@ -0,0 +1,136 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { getQueryString } from '@wikia/ember-fandom/utils/url';
+
+import config from '../config/environment';
+
+// TODO: Remove this when all discussions' posts are in the index
+const QUIZZES_WHITELIST = [
```
rybmat, line 10: I don't have enough context about why this array is needed, but it seems to me like good candidate to replace with some wg variable or if you think that this is overkill then at least to move it to separate file/module together with some function that checks if the request is in context of one of these wikis

app/components/post-search-results.js
```diff
@@ -0,0 +1,136 @@
...
+  wikiVariables: service(),
+
+  isLoading: true,
+  isInternal: false,
+  posts: null,
+  seeMore: false,
+
+  // fortuneatly we can compute the feeds path from articlePath (it has lang part)
+  feedsUrl: computed('wikiVariables.articlePath', function () {
+    return this.get('wikiVariables.articlePath').replace('/wiki/', '/f/');
```
rybmat: is it possible that this code will execute for wikis that do not have feeds/discussions enabled?

app/styles/component/_post-search-results-item.scss
```diff
@@ -0,0 +1,46 @@
...
+  &::after {
+    clear: both;
+    content: "";
+    display: block;
+    height: 0;
+    visibility: hidden;
+  }
+
+  &__title {
+    color: #0d7675;
```
rybmat: is it color not defined in design system? Out of curiosity, can you send me link to designs?

app/templates/components/wikia-ui-components/wikia-card-list.hbs
```diff
@@ -14,5 +14,8 @@
         <div class="wikia-card__snippet wds-font-size-xs">{{item.snippet}}</div>
       {{/link-to}}
     </li>
+    {{#if (equal key 1)}}
+      {{post-search-results query=query seeMore=true isInternal=true}}
+    {{/if}}
```
rybmat, line 21: `wikia-card-list` component is used also in file pages. Is it ok to show `post-search-results` there as well?

app/components/post-search-results.js
vforge: This is very temporary and will be removed in ~2weeks or so.

app/templates/components/wikia-ui-components/wikia-card-list.hbs
vforge: Good catch, I'll check up with Product

app/styles/component/_post-search-results-item.scss
vforge: https://zpl.io/aRzpBkn

app/components/post-search-results.js
vforge: Fixed in e480ec7

app/templates/components/wikia-ui-components/wikia-card-list.hbs
vforge: Fixed in 2b2328f

app/styles/component/_post-search-results-item.scss
vforge: Actually it's awfully close to `$wds-fandom-color-link`, going to use that instead. Fixed in 2b2328f.

app/components/post-search-results.js
vforge: Fixed in 32cc96a

app/components/post-search-results-item.js
vforge: Fixed

app/styles/component/_post-search-results-item.scss
vforge: Actually it's awfully close to `$wds-fandom-color-link`, going to use that instead.

config/bundlesize.js
latata: It's just for monitoring if we do not add too much 💩for example while adding some npm packages

app/components/post-search-results-item.js
```diff
@@ -0,0 +1,25 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+import Thumbnailer from '../modules/thumbnailer';
+
+export default Component.extend({
+  isQuizWithTakes: computed('post.{type,stats,stats.takes}', function () {
```
latata: I think you can just use `post` as a dependency for this computed property. I assume that `post` is not mutated.

app/styles/component/_post-search-results-item.scss
```diff
@@ -0,0 +1,49 @@
...
+    letter-spacing: .03px;
+    line-height: 1;
+    margin-bottom: 9px;
+  }
+
+  &__details {
+    align-items: center;
+    color: $wds-color-dark-gray;
+    display: flex;
+    fill: currentColor;
```
latata: I think it's not needed here. DS `.wds-icon` class already have it.

app/styles/component/_post-search-results.scss
```diff
@@ -0,0 +1,40 @@
+.post-search-results {
+  border-bottom: 1px solid $wds-fandom-color-middle-gray;
+  font-family: $font-family-with-rubik;
```
latata: isn't it set for whole `body`?

app/components/post-search-results.js
```diff
@@ -0,0 +1,150 @@
...
+  wikiVariables: service(),
+
+  isLoading: true,
+  isCrossWiki: false,
+  posts: null,
+  isNotCrossWiki: not('isCrossWiki'),
+
+  // fortunately we can compute the feeds path from articlePath (it has lang part)
+  feedsUrl: computed('wikiVariables.articlePath', function () {
+    return this.get('wikiVariables.articlePath').replace('/wiki/', '/f/');
```
latata: You don't need to use `get` implicitly. 
`this.wikiVariables.articlePath.replace('/wiki/', '/f/')`

vforge: I've seen both approaches in the code. Are there any particular rules then to use `get` and when to omit it?

app/components/post-search-results.js
```diff
@@ -0,0 +1,150 @@
...
+    // otherwise it is enabled
+    return true;
+    */
+  }),
+
+  didInsertElement() {
+    this._super(...arguments);
+
+    if (this.isEnabled) {
+      this.fetchResults(this.get('query'));
```
latata, line 97: 1. what about SSR? isn't it needed here?
2. I'm not an Ember expert but we usually pass data from the route model: https://github.com/Wikia/mobile-wiki/blob/dev/app/routes/search.js#L44

app/components/post-search-results.js
latata: You can use both. They work the same but without `.get()` is shorter :)

app/components/post-search-results.js
vforge: We don't do SSR at all. Also we're going to re-use this module on articles with hand-crafted search query string.

Thanks for the link, I was searching where the param parsing is.

app/styles/component/_post-search-results.scss
vforge: Fixed in 1546f68

app/styles/component/_post-search-results-item.scss
vforge: Fixed in 1546f68

app/components/post-search-results-item.js
vforge: Fixed in 1546f68

app/components/post-search-results.js
latata: Actually there is one difference
`this.something.another` will throw `Cannot read property 'another' of undefined` when `this.something` is undefined
`this.get('something.another')` will just return `undefined`

vforge: Fixed in d3bb4c6

app/components/post-search-results.js
vforge: I'd say that even moving to separate file is an overkill. We intend to enable the new search feature only on quizzes communities at first (because we only index quizzes first) and then after we finish adding all the discussions' posts, we're going to remove this variable completely.

app/styles/component/_post-search-results-item.scss
vforge: This module will go though Design Review after it's merged, so I think we're OK with current colors.

rybmat (approved):

PR #1786, opened 2019-09-16
SER-3436 Display community name for cross-wiki search

mateuszsip:


app/templates/components/wikia-ui-components/wikia-card-crosswiki-list.hbs
```diff
@@ -6,6 +6,7 @@
           {{item.title}}
         </div>
         <div class="wikia-card__snippet wds-font-size-xs">{{item.snippet}}</div>
+        <div class="wikia-card__community_name">{{item.sitename}}</div>
```
slayful: I'm not sure if the font size needs to be specified in the wikia-card__community_name, how about we use `wds-font-size`?

slayful (approved):
Only one suggestion, besides that it LGTM and can be moved straight to QA after addressing.

app/templates/components/wikia-ui-components/wikia-card-crosswiki-list.hbs
mateuszsip: Sure, good catch!

slayful (approved):

PR #1788, opened 2019-09-23
ADEN-9348 Show floor_adhesion close button after some time

dmnsobczak:


app/modules/ads/slots.js
```diff
@@ -379,6 +379,26 @@ export const slots = {
...
+
+    context.push('listeners.slot', {
+      onRenderEnded: (adSlot) => {
+        if (['invisible_high_impact_2', 'floor_adhesion'].includes(adSlot.getSlotName())) {
+          const closeButton = document.querySelector('.out-of-page-template.floor-adhesion .button-close');
+
+          if (closeButton) {
+            closeButton.style.display = 'none';
+
+            setTimeout(() => {
```
nandy-andy, line 394: If `custom.timeToCloseButton.invisible_high_impact_2` or `custom.timeToCloseButton. floor_adhesion` are set to `0` it may not be `0s` (because of the event loop) and they'll be displayed for some `ms` ;) But maybe it's good enough for the experiment 🤷‍♂️

app/modules/ads/slots.js
```diff
@@ -379,6 +379,26 @@ export const slots = {
...
+  setupCloseButtonTimer() {
+    const { context } = window.Wikia.adEngine;
+
+    context.push('listeners.slot', {
+      onRenderEnded: (adSlot) => {
+        if (['invisible_high_impact_2', 'floor_adhesion'].includes(adSlot.getSlotName())) {
+          const closeButton = document.querySelector('.out-of-page-template.floor-adhesion .button-close');
+
+          if (closeButton) {
+            closeButton.style.display = 'none';
```
nandy-andy, line 392: Changing styles in JS may not be a good practice. But having a class in CSS which we switch is probably more bytes ;) And it's an experiment, anyway.

nandy-andy (commented):
LGTM

app/modules/ads/slots.js
dmnsobczak: You right. But anyway this won't be noticeable for the user I guess :)

app/modules/ads/slots.js
dmnsobczak: Absolute right with this.

PR #1795, opened 2019-10-01
[ADEN-9371] Browsi clean up

Loczi94:

### DIFF
--- a/app/modules/ads/index.js
+++ b/app/modules/ads/index.js
@@ -330,20 +330,33 @@ class Ads {
    */
   triggerAfterPageRenderServices() {
     const { bidders } = window.Wikia.adBidders;
-    const { context, slotService } = window.Wikia.adEngine;
+    const { slotService } = window.Wikia.adEngine;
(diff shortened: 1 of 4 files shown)


app/modules/ads/index.js
```diff
@@ -332,36 +331,19 @@ class Ads {
...
-      });
-      this.startAdEngine();
-
-      if (!slotService.getState('top_leaderboard')) {
-        this.finishFirstCall();
-      }
-    } else if (context.get('services.browsi.enabled')) {
-      // Browsi needs googletag loaded
-      this.loadGoogleTag();
-    }
```
bckatarzyna: You can remove only this part of code and leave the rest of 336:364 as it is - in general this refactor with extracting `callExternalTrackingServices` function and checking if ad stack is enabled are ok and still valid so can stay.

bckatarzyna (changes requested):

bckatarzyna (approved):

PR #1799, opened 2019-10-02
CAKE-5177 | Add additional unit and targeting priorities

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5177

## Description

- Add two more properties to unit to drive the looks: `isBig` and `isExternal`
- Add priority to units and use that to determine which unit should be more important to display
- Add `onlyOnAndorid` and `onlyOnIOS` flags
- Add `disableOnSearch` and `disableOnPage` flags

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots-units.json
+++ b/app/services/affiliate-slots-units.json
@@ -1,16 +1,24 @@
 [
   {
     "name": "fandom-com",
+    "isBig": false,
     "image": "https://vignette.wikia.nocookie.net/central/images/b/bc/Wiki.png/revision/latest?cb=20180423162614",
     "heading": "Fandom",
     "subheading": "Check this out",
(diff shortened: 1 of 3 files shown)


app/services/affiliate-slots-units.json
```diff
@@ -1,16 +1,24 @@
 [
   {
     "name": "fandom-com",
+    "isBig": false,
```
mklucsarits, line 4: What is this for?

vforge: it is to difference between when unit is supposed to only be part of the search result vs when unit wants to be alone (ie. the big unit in the mocks).

mklucsarits (approved):

PR #1800, opened 2019-10-02
CAKE-5177 | Cleanup affiliate-slots service

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5177

## Description

- Cleanup the affiliate-slots' logic, remove duplication
- Diff view makes it actually difficult to read

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots.js
+++ b/app/services/affiliate-slots.js
@@ -160,7 +160,7 @@ export default Service.extend({
       // grab only names of allowed units
       .map(t => t.unit)
       // flatten the array
-      .flat()
+      .reduce((prev, curr) => prev.concat(curr))
       // make the list unique
       .filter(distinct);
 
--- a/app/services/affiliate-slots.js
+++ b/app/services/affiliate-slots.js
@@ -5,11 +5,41 @@ import Service, { inject as service } from '@ember/service';
 
 import { system } from '../utils/browser';
 
+/**
+  * @typedef {Object} AffiliateTargeting
+  * @property {string} [comment]
+  * @property {string[]} unit
+  * @property {number} [wikiId=[]]
+  * @property {number} [country=[]]
+  * @property {number|false} [page=[]]
+  * @property {number|false} [query=[]]
+  * @property {number} [vertical=[]]
+*/
 import targeting from './affiliate-slots-targeting';
+
+/**
+  * @typedef {Object} AffiliateUnit
+  * @property {string} name
+  * @property {string} [isBig]
+  * @property {string} image
+  * @property {string} heading
+  * @property {string} subheading
+  * @property {string} link
+  * @property {boolean} isExternal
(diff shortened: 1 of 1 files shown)


JoshuaRogan (approved):

app/services/affiliate-slots.js
```diff
@@ -48,74 +86,93 @@ export default Service.extend({
...
+  getAllUnitsOnSearch(query) {
+    return this.getAllUnits([
+      t => !t.disableOnSearch,
+      t => checkFilter(t.query, query),
+    ]);
   },
 
   /**
-   * Get one random unit from that can be displayed on current wiki with given `query`
+   * Get all units that can be displayed on current wiki with given `title`
```
mklucsarits: This description is wrong; it gets only a single unit.

app/services/affiliate-slots.js
```diff
@@ -5,11 +5,41 @@ import Service, { inject as service } from '@ember/service';
...
 
+/**
+  * @typedef {Object} AffiliateTargeting
+  * @property {string} [comment]
+  * @property {string[]} unit
+  * @property {number} [wikiId=[]]
+  * @property {number} [country=[]]
+  * @property {number|false} [page=[]]
+  * @property {number|false} [query=[]]
+  * @property {number} [vertical=[]]
```
mklucsarits: string, not number

app/services/affiliate-slots.js
vforge: Fixed in 09a2066

app/services/affiliate-slots.js
```diff
@@ -188,7 +188,7 @@ export default Service.extend({
       // grab only names of allowed units
       .map(t => t.unit)
       // flatten the array
-      .reduce((prev, curr) => prev.concat(curr))
+      .reduce((prev, curr) => prev.concat(curr), [])
```
JoshuaRogan, line 191: easy to forget 😓

app/services/affiliate-slots.js
```diff
@@ -86,13 +86,27 @@ export default Service.extend({
   currentCountry: readOnly('geo.country'),
 
   /**
-   * Get one unit that can be displayed on search page with given `query`
+   * Get one BIG unit that can be displayed on search page with given `query`
+   *
+   * @param {string} query
+   * @returns {AffiliateUnit|undefined}
+   */
+  getBigUnitOnSearch(query) {
```
JoshuaRogan, line 94: (feel free to ignore) but maybe we should call it 'Fan Suggested' or something else other than 'Big Unit'

vforge: As this code is supposed (in the future) to handle different units I’d rather stick to big/small than to labels that the UI will have.

PR #1805, opened 2019-10-04
CAKE-5178 Add big affiliate unit to article content

mklucsarits:
## Links
 * http://wikia-inc.atlassian.net/browse/CAKE-5178

## Description
Adds the "big" affiliate unit to the article content

## Reviewers
@Wikia/cake

### DIFF
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -1,0 +1,6 @@
+import Component from '@ember/component';
+
+export default Component.extend(
+  // Intentionally empty, no extra logic needed
+  {},
+);
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -1,6 +1,5 @@
 import Component from '@ember/component';
 
-export default Component.extend(
-  // Intentionally empty, no extra logic needed
-  {},
-);
+export default Component.extend({
+  classNames: ['affiliate-unit'],
+});
--- a/app/styles/component/_affiliate-unit.scss
+++ b/app/styles/component/_affiliate-unit.scss
@@ -1,6 +1,14 @@
 .affiliate-unit {
+  margin: 44px auto;
+
+  &__header-icon {
+    fill: $wds-fandom-color-navy;
+  }
+
   &__header {
-    font-weight: $wds-font-weight-medium;
+    color: $wds-fandom-color-navy;
+    margin-bottom: 20px;
+    text-align: center;
     text-transform: uppercase;
   }
 
@@ -24,7 +32,7 @@
     height: 100%;
     justify-content: flex-end;
     left: 0;
-    padding: 18px 14px;
+    padding: 14px 18px;
     position: absolute;
     top: 0;
     width: 100%;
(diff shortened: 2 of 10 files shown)


app/components/affiliate-unit.js
```diff
@@ -0,0 +1,6 @@
+import Component from '@ember/component';
+
+export default Component.extend(
+  // Intentionally empty, no extra logic needed
```
JoshuaRogan: Is this required if we don't need it ?

JoshuaRogan (approved):

app/components/affiliate-unit.js
mklucsarits: Yes, an error occurs if the component file doesn't exist.

vforge: :emberjs:

app/templates/components/affiliate-unit.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="affiliate-unit__header">Suggested By Fans</div>
```
vforge: Please add this to i18n.

app/styles/component/index.scss
```diff
@@ -31,6 +31,7 @@
 @import 'avatar';
 @import 'language-wikis-index';
 @import 'watch-show';
+@import 'affiliate-unit';
```
vforge: Can this file be lazy loaded?

app/styles/component/_affiliate-unit.scss
```diff
@@ -0,0 +1,43 @@
...
+    justify-content: flex-end;
+    left: 0;
+    padding: 18px 14px;
+    position: absolute;
+    top: 0;
+    width: 100%;
+  }
+
+  &__heading {
+    color: #fff;
```
vforge: Please use scss var for the color

app/styles/component/_affiliate-unit.scss
```diff
@@ -0,0 +1,43 @@
...
+    font-weight: bold;
```
vforge: Instead of using scss for typography mobile-wiki seems to be using class names - ie. `.wds-font-weight-bold` and `.wds-leading-tight`

https://fandomdesignsystem.com/base-styles/typography

app/templates/components/affiliate-unit.hbs
mklucsarits: I was rather thinking this text might become part of our slots configuration in the JSON. But in the meantime I can add it to i18n, yes.

app/styles/component/index.scss
mklucsarits: I'm not sure that makes sense if the HTML is injected straight-away. But if it was lazy loaded, how do you propose to do that? I've never seen SCSS lazy loaded as a pattern in any of our apps.

app/styles/component/_affiliate-unit.scss
mklucsarits: Even for white? 👎

vforge (approved):

PR #1821, opened 2019-10-09
ADEN-8905 Track adengine on/off.

jacekbj:
This ticket introduces a fundamental change into how AdEngine is loaded in Mobile-Wiki.

The main point is `routes/wiki-page.js`. This is the decision point for loading AdEngine.

If any part of Mobile-Wiki requires AdEngine loaded, it needs to use `Ads.getLoadedInstance()` method. It returns a Promise which:
* resolves when AdEngine is loaded
* rejects when a decision is made not to load AdEngine or AdEngine assets fail to load.

This way there is no risk of pending promises hanging on backend.


app/modules/ads/index.js
```diff
@@ -154,11 +187,6 @@ class Ads {
   startAdEngine() {
     const { AdEngine } = window.Wikia.adEngine;
 
-    if (!this.isAdStackEnabled()) {
```
fraszczakszymon, line 157: Why is this removed? Smart banner relies on `fanTakeoverResolver` decision. Does it work properly now?

jacekbj: https://github.com/Wikia/mobile-wiki/blob/8b630f7e9a1f1678456a06ba402669cf07ef3ace/app/modules/ads/index.js#L336

`startAdEngine` is only run if `this.isAdStackEnabled()` is equal to true. Therefore this condition will never be fulfilled.

`fanTakeoverResolver` is resolved here: https://github.com/Wikia/mobile-wiki/blob/8b630f7e9a1f1678456a06ba402669cf07ef3ace/app/modules/ads/index.js#L184

Truth be told, I did not test this scenario when writing code.

fraszczakszymon: :facepalm:

Does it mean that smart banner is broken for some time already? It seems that fanTakeoverResolved isn't resolved at all when noads/ad stack is disabled :(

jacekbj: I fixed smart banner by resolving the promise if ad stack is disabled.

app/controllers/application.js
```diff
@@ -44,10 +38,11 @@ export default Controller.extend(
...
 
     file: alias('lightbox.file'),
-    fullPage: oneWay('mobileApp'),
+    fullPage: oneWay('wikiPage.mobileApp'),
     isSearchPage: equal('currentRouteName', 'search'),
-    noAds: alias('ads.noAdsQueryParam'),
-    mobileApp: alias('ads.disableAdsInMobileApp'),
+    mobileApp: oneWay('wikiPage.mobileApp'),
+
+    ads: Ads.getInstance(),
```
fraszczakszymon: Seems not needed, is it?

app/modules/ads/index.js
```diff
@@ -179,20 +142,14 @@ class Ads {
...
 
   createJWPlayerVideoAds(options) {
     const { jwplayerAdsFactory } = window.Wikia.adProducts;
 
-    if (this.isAdStackEnabled()) {
-      return jwplayerAdsFactory.create(options);
-    }
-
-    return null;
+    return jwplayerAdsFactory.create(options);
```
fraszczakszymon, line 155: What if there is no ad stack on FV page?

app/modules/ads/index.js
```diff
@@ -46,16 +44,10 @@ class Ads {
       return adsPromise;
     }
 
-    if (typeof FastBoot !== 'undefined') {
```
fraszczakszymon, line 49: It's causing a memory leak on the server side, isn't it? Am I missing something?

app/components/ad-slot.js
```diff
@@ -64,10 +57,6 @@ export default Component.extend(
       const ads = this.get('ads.module');
       const name = this.name;
 
-      if (this.noAds) {
-        return;
-      }
-
       if (this.insertOnViewportEnter) {
         ads.pushSlotToQueue(name);
```
fraszczakszymon: Hihi, this part doesn't make sense, right?
https://github.com/Wikia/mobile-wiki/pull/1821/files#diff-98564636413dff58d77852057b876c48R76-R77

app/modules/ads/index.js
```diff
@@ -179,20 +142,14 @@ class Ads {
   finishFirstCall() {
     const { btfBlockerService } = window.Wikia.adEngine;
 
-    if (this.isAdStackEnabled()) {
-      btfBlockerService.finishFirstCall();
-      fanTakeoverResolver.resolve();
-    }
+    btfBlockerService.finishFirstCall();
```
fraszczakszymon, line 145: What if there is no ad stack on page?

fraszczakszymon: Shouldn't we mark it as private? (if it's executed in safe place)

app/modules/ads/index.js
```diff
@@ -547,10 +489,6 @@ class Ads {
   waitForVideoBidders() {
     const { context, utils } = window.Wikia.adEngine;
 
-    if (!this.isAdStackEnabled()) {
```
fraszczakszymon, line 564: What if there is no ad stack? FV promise may not be resolved?

fraszczakszymon: Never mind, I've noticed the changes in `video-players/jwplayer.js`

app/modules/ads/index.js
fraszczakszymon: Never mind, I've noticed your changes: 
https://github.com/Wikia/mobile-wiki/pull/1821/files#diff-1f8a4e7742e20ee4da173fdc8f902a2fR71

app/components/ad-slot.js
```diff
@@ -40,10 +37,6 @@ export default Component.extend(
         return;
       }
 
-      if (this.noAds) {
```
fraszczakszymon, line 43: Do we need to push slot when ad stack is disabled?

app/modules/ads/index.js
fraszczakszymon: Thank you

app/components/ad-slot.js
jacekbj: Here I made sure that if ads are disabled then `Ads.waitForAdEngine` will not resolve
https://github.com/Wikia/mobile-wiki/pull/1821/commits/6c3ce1973e1d13a325e890b55e90b3e83608c1fc

Ads are injected after `Ads.waitForAdEngine` resolves, therefore there is no need to check for that in component:
https://github.com/Wikia/mobile-wiki/blob/ADEN-8905/app/components/article-content.js

app/components/ad-slot.js
jacekbj: That's right, I removed `ads` service in favour of `Ads` module.
https://github.com/Wikia/mobile-wiki/pull/1821/commits/25acef63cb57ce43676d7621c6f965a638f071a4

app/controllers/application.js
jacekbj: Removed.

app/modules/ads/index.js
jacekbj: There should be no memory leaks now.

app/modules/ads/index.js
jacekbj: Marked.

app/components/ad-slot.js
fraszczakszymon: :+1:

app/components/ad-slot.js
```diff
@@ -61,41 +52,41 @@ export default Component.extend(
...
-      } else if (this.shouldWaitForUapResponse) {
-        ads.waitForUapResponse(
-          () => ads.pushSlotToQueue(this.name),
-          () => {},
-        );
-      }
+      Ads.getLoadedInstance()
+        .then((ads) => {
+          if (this.insertOnViewportEnter) {
+            ads.pushSlotToQueue(this.name);
```
fraszczakszymon, line 58: `pushSlotToQueue` does not take any argument. Also, this if statement is broken (?). In this case, when `insertOnViewportEnter` is truthy, method `pushSlotToQueue` does not take any actions (returns void when `insertOnViewportEnter` is truthy). 🙈

fraszczakszymon: https://github.com/Wikia/mobile-wiki/pull/1821/files#diff-98564636413dff58d77852057b876c48R75

fraszczakszymon (approved):

app/components/ad-slot.js
jacekbj: I tried to refactor this part but then got kind of dizzy looking at how convoluted this is. I will straighten this up.

jacekbj: My attempt to clean this place up deserves a ticket on its own.
Leaving it as is for now.

PR #1824, opened 2019-10-09
CAKE-5209 Design updates and fixes for search results interrupt

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5209

## Description
 * Fixed styling issues to match mocked up design
 * Removed extraneous div elements around components

## Reviewers
@Wikia/cake

### DIFF
--- a/app/components/post-search-results-affiliate-item.js
+++ b/app/components/post-search-results-affiliate-item.js
@@ -3,6 +3,7 @@ import Component from '@ember/component';
 import { track, trackActions } from '../utils/track';
 
 export default Component.extend({
+  classNames: ['post-search-results-affiliate-item'],
   actions: {
     trackClick(number) {
       track({
--- a/app/components/post-search-results-item.js
+++ b/app/components/post-search-results-item.js
@@ -5,6 +5,7 @@ import { track, trackActions } from '../utils/track';
 import Thumbnailer from '../modules/thumbnailer';
 
 export default Component.extend({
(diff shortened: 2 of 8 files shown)


vforge (approved):

app/components/post-search-results-affiliate-item.js
```diff
@@ -3,6 +3,7 @@ import Component from '@ember/component';
 import { track, trackActions } from '../utils/track';
 
 export default Component.extend({
+  classNames: ['post-search-results-affiliate-item'],
```
JoshuaRogan, line 6: Interesting. Does that mean there is always a div for each component?

vforge: Yeah, ember takes care of it

vforge:
```

mobile-wiki/templates/components/post-search-results.hbs

  32:6  error  Incorrect indentation for `{{#each}}` beginning at L32:C6. Expected `{{#each}}` to be at an indentation of 8 but was found at 6.  block-indentation
```

app/components/post-search-results-affiliate-item.js
JoshuaRogan: Hmmm. I'm not sure if I like that or find it to be wasteful if you have the template.

vforge: Regardless it will always* be there.

PR #1827, opened 2019-10-10
CAKE-5208 Fix one design bug and update test targeting

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5208

## Description
 * Show big affiliate unit when searching for "test" on xkxd wiki -- it won't show with the existing terms because there are no results.
 * Add a border-radius for the element containing the image.

## Reviewers
@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots-targeting.json
+++ b/app/services/affiliate-slots-targeting.json
@@ -33,6 +33,8 @@
       1619010
     ],
     "query": [
+      "spiderman",
+      "marvel",
       "test"
     ]
   },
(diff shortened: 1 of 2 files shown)


JoshuaRogan (approved):

vforge (approved):

app/services/affiliate-slots-targeting.json
```diff
@@ -33,8 +33,7 @@
       1619010
     ],
     "query": [
-      "spiderman",
-      "marvel"
```
vforge, line 37: Can we not remove those? https://wikia-inc.atlassian.net/browse/CAKE-5208?focusedCommentId=470463&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-470463

PR #1832, opened 2019-10-11
CAKE-5199 | Twak Ad insert logic

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5199

## Description

Based on all the comments in https://wikia-inc.atlassian.net/browse/ADEN-9465 I created this solution. It might be improved later by AdEng

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/ads/ad-slot-builder.js
+++ b/app/services/ads/ad-slot-builder.js
@@ -4,6 +4,21 @@ import { getRenderComponentFor } from '../../utils/render-component';
 
 const MIN_ZEROTH_SECTION_LENGTH = 700;
 
+/**
+ * Check if element is too close `.affiliate-slot`
+ *
+ * @param {DOMElement} element
+ * @returns {boolean}
+ */
+function tooCloseToAffiliateSlots(element) {
+  const elementTop = offset(element).top || 0;
+
+  return Array.prototype.slice
+    .call(element.parentNode.querySelectorAll('.article-content > .affiliate-slot'))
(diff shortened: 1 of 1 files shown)


mklucsarits (approved):
In theory I like the idea, but I don't know how to test it locally.

app/services/ads/ad-slot-builder.js
```diff
@@ -4,6 +4,21 @@ import { getRenderComponentFor } from '../../utils/render-component';
...
+ * @param {DOMElement} element
+ * @returns {boolean}
+ */
+function tooCloseToAffiliateSlots(element) {
+  const elementTop = offset(element).top || 0;
+
+  return Array.prototype.slice
+    .call(element.parentNode.querySelectorAll('.article-content > .affiliate-slot'))
+    .map(e => (offset(e).top || 0))
+    .some(top => (Math.abs(top - elementTop) < MIN_ZEROTH_SECTION_LENGTH));
```
jacekbj: I would suggest using a different variable name here. `MIN_ZEROTH_SECTION_LENGTH` is used to control the distance of the first boxad from the top of the page. In this function the variable controls the distance from top of affiliate slot to h2 element. What do you think about `AFFILIATE_SLOT_HEIGHT` or sth similar?

jacekbj (approved):

app/services/ads/ad-slot-builder.js
vforge: 1f36d46

PR #1840, opened 2019-10-17
ADEN-9420 Added krux segment tracking to DW

preludek:
## Links

* more relevant links

## Description

## Reviewers

### DIFF
--- a/app/modules/ads/index.js
+++ b/app/modules/ads/index.js
@@ -419,9 +419,9 @@ class Ads {
   trackKruxSegments() {
     const { context, utils } = window.Wikia.adEngine;
     const kruxUserSegments = context.get('targeting.ksg') || [];
-    const kruxICBMSegments = context.get('services.krux.icbmSegments') || [];
+    const kruxTrackedSegments = context.get('services.krux.trackedSegments') || [];
 
-    const kruxPropValue = kruxUserSegments.filter(segment => kruxICBMSegments.includes(segment));
+    const kruxPropValue = kruxUserSegments.filter(segment => kruxTrackedSegments.includes(segment));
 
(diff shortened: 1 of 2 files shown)


app/modules/ads/index.js
```diff
@@ -413,6 +413,22 @@ class Ads {
     }
   }
 
+  /**
+   * @private
+   */
+  trackKruxSegments() {
+    const { context, utils } = window.Wikia.adEngine;
+    const kruxUserSegments = context.get('targeting.ksg') || [];
+    const kruxICBMSegments = context.get('services.krux.icbmSegments') || [];
```
dmnsobczak: Can we rename it to `kruxTrackedSegments` and `services.krux.trackedSegments`? I don't see the point of using ICBM in the variable names since almost all context values comes from ICBM.

app/modules/ads/index.js
```diff
@@ -413,6 +413,22 @@ class Ads {
...
+
+    const kruxPropValue = kruxUserSegments.filter(segment => kruxICBMSegments.includes(segment));
+
+    if (kruxPropValue.length) {
```
dmnsobczak, line 426: AC: `empty string if in no segments from list`. Something changed?

dmnsobczak (changes requested):
Please confirm that AC has changed.

app/modules/ads/index.js
preludek: ✅

app/modules/ads/index.js
preludek: same conclusion as for desktop

dmnsobczak (approved):

PR #1841, opened 2019-10-18
[ADEN-9376] x-click tracking

Loczi94:
X clicks are tracked to GA.


dmnsobczak (changes requested):
`not ok 8 - jwplayer:js: 55.1KB > 55KB (uncompressed)`
`Bundlesize check failed with 1 error!`

dmnsobczak (approved):

PR #1870, opened 2019-11-05
CAKE-5248 Disney+ affiliate ad targeting

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5248

## Description
Use production-ready json data for Disney+ affiliate ad targeting.

## Reviewers
@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots-targeting.json
+++ b/app/services/affiliate-slots-targeting.json
@@ -1,43 +1,74 @@
 [
   {
-    "campaign": "fandom-test",
-    "category": "fandom-test",
+    "campaign": "disneyplus",
+    "category": "disney",
     "wikiId": [
-      1081891
-    ],
-    "query": [
-      "coffee"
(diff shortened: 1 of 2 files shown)


app/services/affiliate-slots-targeting.json
```diff
@@ -1,43 +1,74 @@
...
-    "campaign": "fandom-test",
-    "category": "fandom-test",
+    "campaign": "disneyplus",
+    "category": "disney",
     "wikiId": [
-      1081891
-    ],
-    "query": [
-      "coffee"
+      374,
```
vforge: Are all those for the same set of wiki ids? Then the first one will ever be taken - we should limit either wikis OR add query so more diverisive set of units are available.

JoshuaRogan (approved):

vforge (approved):

PR #1872, opened 2019-11-06
CAKE-5234 Fix call to nonexistent method

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5234

## Description
The existing code didn't account for showing the disclaimer in the case of a search result affiliate unit appearing on the article page. This resulted in an error because the method wasn't passed to the component attributes.

@Wikia/cake


cschucode (approved):
LGTM.

vforge (approved):

PR #1873, opened 2019-11-06
Cake 5273 | Design Modification to Watch Now Module

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5273
* parent ticket: http://wikia-inc.atlassian.net/browse/CAKE-5251

## Description

- Move the call to `handleWatchShow` to display after infoboxes
- Add design slant to "Watch Now" module
- Style

## Reviewers

@Wikia/cake


mklucsarits (changes requested):
I don't think this is an ideal solution. We should figure out why the code as written is placing the ad above "watch show" when it's supposed to load it outside/after `.portable-infobox-wrapper` (https://github.com/Wikia/mobile-wiki/blob/dev/app/services/ads/ad-slot-builder.js#L36). Additionally, waiting 5 seconds is an eternity in the mobile web world. If the ad doesn't load because the user is logged in, there's a good chance they'll have already scrolled the page and won't see "watch show" at all by the time it loads.

cschucode:
I agree and after another look, it appears that the ordering of the `handleWatchShow` can be moved down to achieve the desired effect without messing with or relying on ad code.

mklucsarits (approved):

mklucsarits (approved):

PR #1876, opened 2019-11-06
CAKE-5251 Fix placement of 'watch show' in DOM

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5251

## Description
Places the "watch show" promo unit where it should be, to prevent the top leaderboard ad from appearing above it. With the existing logic, for whatever reason, Ember inserts the "watch show" component inside the top leaderboard container. With this fix, we instead wait until the infobox is replaced with the Ember component, which includes the `.portable-infobox-wrapper`. Then we put the "watch show" component inside that.

## Reviewers
@Wikia/cake

### DIFF
--- a/app/components/article-content.js
+++ b/app/components/article-content.js
@@ -64,9 +64,9 @@ export default Component.extend(
 
           this.handleBigAffiliateUnit();
           this.handlePostSearchResults();
-          this.handleWatchShow();
           this.handleInfoboxes();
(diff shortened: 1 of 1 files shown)


mklucsarits:
Copied into another PR (https://github.com/Wikia/mobile-wiki/pull/1873)

PR #1878, opened 2019-11-07
Enable post-search-results everywhere

vforge:
## Description

- Enable post-search-results everywhere*
- Fix escaped `&hellip;`
- Fix styles for "SEE MORE ->"

*) if the community has discussions enabled OR this is the cross wiki search

## Reviewers

@Wikia/cake

### DIFF
--- a/app/styles/component/_post-search-results.scss
+++ b/app/styles/component/_post-search-results.scss
@@ -4,6 +4,7 @@
   padding-bottom: 10px;
 
   &__header {
+    display: flex;
     margin-bottom: 14px;
     text-align: center;
   }
--- a/app/templates/components/post-search-results-item.hbs
+++ b/app/templates/components/post-search-results-item.hbs
@@ -6,7 +6,7 @@
   class="post-search-results-item__title wds-leading-tight"
   onclick={{action "trackClick" postIndex}}
 >
-  {{getTitle}}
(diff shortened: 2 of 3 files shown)


JoshuaRogan (approved):

app/styles/component/_post-search-results.scss
```diff
@@ -4,6 +4,7 @@
   padding-bottom: 10px;
 
   &__header {
+    display: flex;
```
JoshuaRogan, line 7: Any support issues with this?

app/templates/components/post-search-results-item.hbs
```diff
@@ -6,7 +6,7 @@
   class="post-search-results-item__title wds-leading-tight"
   onclick={{action "trackClick" postIndex}}
 >
-  {{getTitle}}
+  {{{getTitle}}}
```
JoshuaRogan, line 9: Why the extra `{}`?

app/styles/component/_post-search-results.scss
vforge: nothing that we should care about :)

app/templates/components/post-search-results-item.hbs
vforge: un-escape the string

PR #1885, opened 2019-11-08
ADEN-9593 fix double icb1 creation

preludek:
## Links

* more relevant links

## Description

## Reviewers

### DIFF
--- a/app/modules/ads/slots-loader.js
+++ b/app/modules/ads/slots-loader.js
@@ -129,11 +129,14 @@ export const slotsLoader = {
   },
 
   loadFirstSlot() {
+    const { context } = window.Wikia.adEngine;
     const firstSlotName = `${this.baseSlotName}_1`;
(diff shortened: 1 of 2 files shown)


app/modules/ads/slots-loader.js
```diff
@@ -129,11 +129,14 @@ export const slotsLoader = {
   },
 
   loadFirstSlot() {
+    const { context } = window.Wikia.adEngine;
     const firstSlotName = `${this.baseSlotName}_1`;
+    const isDisabledOnScroll = context.get('slots.incontent_boxad_1.repeat.disablePushOnScroll');
```
fraszczakszymon: `loadFirstSlot` method seems to be general, it isn't used by `incontent_boxad_1` only. Why do we look for value on this particular slot here?

preludek: it's the only repeatable slot atm and the function always calls for icb1. Still a very good point so it will be generalized

fraszczakszymon (approved):

PR #1897, opened 2019-11-12
[hotfix] Reduce height of margin below 'Watch Show' module

mklucsarits:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5276

## Description
Reduce spacing below the 'Watch Show' module.


vforge (approved):

cschucode (commented):
I would double-check this one with design. We set that a little high so the linear-gradient would get a smooth blend into the background.

mklucsarits:
This messes up the gradient that creates the slanted line. It'll have to be fixed in a more careful way.

PR #1910, opened 2019-11-14
IW | pass env property to fetch service config

rybmat:
## Description

Pass environment to fetch service config. Without that, fastboot communicates with pandora services using ingress urls instead of k8s internal ones

## Reviewers

@Wikia/iwing


app/services/affiliate-slots.js
```diff
@@ -1,4 +1,3 @@
-import { computed } from '@ember/object';
```
rybmat, line 1: unused one

bkoval (approved):

PR #1914, opened 2019-11-14
CAKE-5299 | Only client-side render for interrupt

vforge:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5299

## Description

- Use `didInsertElement` for interrupt, thus disabling interrupt in fastboot

## Reviewers

@Wikia/cake

### DIFF
--- a/app/components/post-search-interrupt.js
+++ b/app/components/post-search-interrupt.js
@@ -13,9 +13,8 @@ export default Component.extend({
 
   isEnabled: and('isBigFinished', 'isSmallFinished'),
 
-  init() {
+  didInsertElement() {
     this._super(...arguments);
-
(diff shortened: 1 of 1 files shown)


app/components/post-search-interrupt.js
```diff
@@ -13,10 +13,9 @@ export default Component.extend({
 
   isEnabled: and('isBigFinished', 'isSmallFinished'),
 
-  init() {
+  didInsertElement() {
```
JoshuaRogan, line 12: Ahhhh that makes sense

JoshuaRogan (approved):

mklucsarits (approved):

PR #1920, opened 2019-11-15
cake-5305 | Update disclaimer styles on affiliate units

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5305

## Description

These are the updated disclaimer styles for:
- Big Unit
- Post Search Results
- Search Results

## Reviewers

@Wikia/cake

### DIFF
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -10,6 +10,8 @@ export default Component.extend(
 
     classNames: ['affiliate-unit'],
 
+    showAffiiateUnitDisclaimer: document.querySelector('.watch-show__disclaimer') ? false : true,
+
     actions: {
       trackAffiliateClick() {
         trackAffiliateUnit(this.unit, {
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -2,7 +2,6 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import InViewportMixin from 'ember-in-viewport';
-import exists from '../utils/exists';
 import { trackAffiliateUnit, trackActions } from '../utils/track';
 
 export default Component.extend(
@@ -15,7 +14,7 @@ export default Component.extend(
 
     classNames: ['affiliate-unit'],
 
-    showAffiiateUnitDisclaimer: !exists('.watch-show__disclaimer'),
+    showAffiiateUnitDisclaimer: !document.querySelector('.watch-show__disclaimer'),
 
     heading: computed('unit', function () {
       if (this.unit && this.unit.tagline) {
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -1,5 +1,6 @@
 import Component from '@ember/component';
 import InViewportMixin from 'ember-in-viewport';
+import exists from '../utils/exists';
 import { trackAffiliateUnit, trackActions } from '../utils/track';
 
 export default Component.extend(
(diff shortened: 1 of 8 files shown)


vforge:
Quick question: How does it connect to CAKE-5293? If I read it correctly there should be only one text.

mklucsarits (approved):

cschucode:
@vforge this ticket is only styling the units and [CAKE-5302](https://wikia-inc.atlassian.net/secure/RapidBoard.jspa?rapidView=347&modal=detail&selectedIssue=CAKE-5302) will be responsible for when to show the disclaimer.

app/styles/component/_post-search-results.scss
```diff
@@ -40,6 +40,24 @@
     justify-content: center;
     margin: 18px;
   }
+
+  &__disclaimer {
+    color: $wds-fandom-color-dark-gray;
+    font-size: $wds-font-size-xxs;
+    font-style: italic;
+    letter-spacing: .03px;
```
mklucsarits: Does this amount even make a visual difference? I wonder if this kind of stuff is intentional or just a weird byproduct of importing into Zeplin.

mklucsarits (approved):

app/styles/component/_post-search-results.scss
cschucode: I did notice a change and wanted to keep things up to spec for the designers. But I know what you mean.

app/components/affiliate-unit.js
```diff
@@ -14,6 +15,8 @@ export default Component.extend(
 
     classNames: ['affiliate-unit'],
 
+    showAffiiateUnitDisclaimer: !exists('.watch-show__disclaimer'),
```
JoshuaRogan: When will this value be evaluated? Is it a possible race condition between the loading of the JS and this code?

app/utils/exists.js
```diff
@@ -0,0 +1,4 @@
+export default function exists(selector) {
+  const el = document.querySelector(selector);
+  return !!el;
```
JoshuaRogan: This can be one line: `return !!document.querySelector(selector);` Which begs the question of needing it to be it's own function.

app/components/affiliate-unit.js
cschucode: Are you asking when will `exists()` be evaluated? This was tested and worked in both dev and sandbox environments. Any edge cases I'm not thinking of?

app/utils/exists.js
cschucode: This was actually to help keep track of the back and forth logic between when to show and hide certain disclaimers. I was also able to reuse in `post-search-results.js`.

app/styles/component/_post-search-results.scss
cschucode: Will double check with design just to make sure.

cschucode: Looks like it was a cross-app translation error between Figma and Zeplin. The correct `letter-spacing` should be `.25rem`.

app/utils/exists.js
cschucode: Looks cleaner: 325d13f

app/components/affiliate-unit.js
JoshuaRogan: Did you test it on a slow connection? I'm wondering if there is a race condition where it evaluates this property before it shows up on the page.

cschucode: Tested both the big affiliate unit and post search results affiliate unit in Chrome under Slow 3G conditions and disclaimer gets rendered with the unit.

JoshuaRogan (approved):

PR #1927, opened 2019-11-18
[ADEN-9528] Send comics=1 page level key-val to GAM

Loczi94:
## Links

https://wikia-inc.atlassian.net/browse/ADEN-9528


app/modules/ads/index.js
```diff
@@ -371,6 +371,9 @@ class Ads {
   startAdEngine() {
     const { AdEngine } = window.Wikia.adEngine;
 
+    const { taxonomyService } = window.Wikia.adServices;
```
preludek: we should move this import up a line so all imports are grouped together

app/modules/ads/index.js
```diff
@@ -371,6 +371,9 @@ class Ads {
...
+    taxonomyService.configureComicsTargeting();
```
preludek: is startAdEngine the correct place for that?

app/modules/ads/setup.js
```diff
@@ -179,6 +179,12 @@ export const adsSetup = {
     context.set('services.moatYi.enabled', instantConfig.isGeoEnabled('wgAdDriverMoatYieldIntelligenceCountries'));
     context.set('services.nielsen.enabled', instantConfig.isGeoEnabled('wgAdDriverNielsenCountries'));
 
+    if (instantConfig.get('icTaxonomyComicsTag')) {
+      context.set('services.taxonomy.comics.enabled', instantConfig.get('icTaxonomyComicsTag'));
```
preludek: From my understanding we later check only if this flag is true so maybe just set it to true? If instantConfig.get('icTaxonomyComicsTag') is not true we won't get here anyway.

app/modules/ads/index.js
preludek: we should move this import up a line so all imports are grouped together

app/modules/ads/index.js
preludek: is startAdEngine the correct place for that? Maybe move it with rest of the tracking ->  triggerPageTracking

app/modules/ads/setup.js
preludek: From my understanding we later check only if this flag is true so maybe just set it to true? If instantConfig.get('icTaxonomyComicsTag') is not true we won't get here anyway.

preludek (changes requested):

app/modules/ads/setup.js
Loczi94: Yep, make sense.

preludek (approved):

PR #1929, opened 2019-11-18
IW-2576 | 💙💚 deployment / get rid of ceph usage

latata:
## Links

* http://wikia-inc.atlassian.net/browse/IW-2576


rybmat (approved):
There should also be Dockerfile for s3cmd image, we should remove it

Jenkinsfile-build-and-deploy
```diff
@@ -1,12 +1,16 @@
+import groovy.json.JsonSlurperClassic
+
 def deployenv = 'artifactory.wikia-inc.com/platform/alpine:3.6-curl'
-def kubectl = 'artifactory.wikia-inc.com/ops/k8s-deployer:0.0.21'
+def kubectl = 'artifactory.wikia-inc.com/ops/k8s-deployer:0.0.22'
```
mszabo-wikia, line 4: There should be an env var available with latest deployer version

mszabo-wikia (approved):

PR #1930, opened 2019-11-18
CAKE-5314 | Add preferred index

JoshuaRogan:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5314

## Description

Use preferred index = 0 for all small units. 

## Reviewers
@Wikia/cake

### DIFF
--- a/app/components/post-search-results.js
+++ b/app/components/post-search-results.js
@@ -11,13 +11,17 @@ import extend from '../utils/extend';
 const DEFAULT_AFFILIATE_SLOT = 1;
 
 function getAffiliateSlot(smallAffiliateUnit, posts) {
+  let preferredIndex = DEFAULT_AFFILIATE_SLOT;
+
   if (!posts || posts.length === 0) {
     return 0;
   }
 
-  const preferredIndex = smallAffiliateUnit.preferredIndex || DEFAULT_AFFILIATE_SLOT;
+  if (smallAffiliateUnit.preferredIndex !== undefined) {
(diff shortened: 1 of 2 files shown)


app/components/post-search-results.js
```diff
@@ -11,13 +11,17 @@ import extend from '../utils/extend';
...
     return 0;
   }
 
-  const preferredIndex = smallAffiliateUnit.preferredIndex || DEFAULT_AFFILIATE_SLOT;
+  if (smallAffiliateUnit.preferredIndex !== undefined) {
+    preferredIndex = smallAffiliateUnit.preferredIndex;
+  }
 
-  if (preferredIndex < posts.length) {
+  if (preferredIndex > posts.length) {
```
mklucsarits, line 24: ;)

mklucsarits (approved):
LGTM, although I'm wondering if we should just make 0 the new "default" index instead of even editing the JSON. 🤷‍♂

app/components/post-search-results.js
JoshuaRogan: No one to blame but myself for this error. 🤦‍♂

JoshuaRogan:
> LGTM, although I'm wondering if we should just make 0 the new "default" index instead of even editing the JSON. 🤷‍♂

Good suggestion https://github.com/Wikia/mobile-wiki/pull/1930/commits/18d9d8608696f3564357c543186810f94461aa7b

mklucsarits (approved):

PR #1938, opened 2019-11-21
cake-5318 | Add logic for rendering updated affiliate links

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5318
* [Updated Links](https://docs.google.com/spreadsheets/d/1vGuJEZ0ncVQSXTrABM3YYHg2P531sWZosgNsq-uMy5M/edit#gid=1139825318)

## Description
For tracking purposes, we want to use affiliate links by unit and page for
- In-content Big Unit
- In-content Small Unit
- Search Big Unit
- Search Small Unit

In the code
- Updated `affiliate-slots-units.json` with affiliate unit links by page and position
- Added logic to display appropriate affiliate unit link
- Removed duplicate units and added "AU" and "NZ" to existing ones

## Reviewers
@Wikia/cake

### DIFF
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -23,15 +23,13 @@ export default Component.extend(
       return this.i18n.t('affiliate-unit.big-unit-heading');
     }),
 
-    getUnitLink: computed('link', function () {
-      let unitLink = this.unit.link;
-
-      if (this.isInContent && this.unit.links.page) {
-        unitLink = this.unit.links.page;
-      } else if (this.unit.links.search) {
-        unitLink = this.unit.links.search;
+    getUnitLink: computed('unit', 'isInContent', function () {
+      if (this.isInContent && this.unit.links && this.unit.links.article) {
+        this.unit.link = this.unit.links.article;
+      } else if (this.unit.links && this.unit.links.search) {
+        this.unit.link = this.unit.links.search;
       }
-      return unitLink;
+      return this.unit.link;
     }),
 
     actions: {
--- a/app/components/affiliate-unit.js
+++ b/app/components/affiliate-unit.js
@@ -23,6 +23,17 @@ export default Component.extend(
       return this.i18n.t('affiliate-unit.big-unit-heading');
     }),
 
+    getUnitLink: computed('link', function () {
+      let unitLink = this.unit.link;
+
+      if (this.isInContent && this.unit.links.page) {
+        unitLink = this.unit.links.page;
+      } else if (this.unit.links.search) {
+        unitLink = this.unit.links.search;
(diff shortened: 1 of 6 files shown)


app/components/affiliate-unit.js
```diff
@@ -23,6 +23,17 @@ export default Component.extend(
       return this.i18n.t('affiliate-unit.big-unit-heading');
     }),
 
+    getUnitLink: computed('link', function () {
```
vforge: Instead of `link` this should listen on both `isInContent` and `unit`.

app/components/affiliate-unit.js
```diff
@@ -23,6 +23,17 @@ export default Component.extend(
...
+      let unitLink = this.unit.link;
+
+      if (this.isInContent && this.unit.links.page) {
```
vforge: Make sure you check if `unit.links` exists (same below)

app/components/post-search-results-affiliate-item.js
```diff
@@ -7,6 +8,17 @@ export default Component.extend({
 
   classNames: ['post-search-results-affiliate-item'],
 
+  getUnitLink: computed('link', function () {
```
vforge: Instead of `link` this should listen on both `isInContent` and `affiliateUnit `.

app/services/affiliate-slots-units.json
```diff
@@ -28,6 +28,10 @@
     "heading": "Moana, a heroine kids and adults alike will admire and cheer for",
     "subheading": "Watch Now On Disney+",
     "link": "https://disneyplus.bn5x.net/DrrRa",
+    "links": {
+      "page": "https://disneyplus.bn5x.net/qq5qq",
```
mklucsarits: I think "article" might be a better name, "page" could be any page, even the search page.

cschucode: agreed. f267051

app/components/affiliate-unit.js
cschucode: Also, had to throw in a check for `!this.isInContent`.

config/bundlesize.js
```diff
@@ -5,7 +5,7 @@ const assetsFolder = 'mobile-wiki/assets';
 module.exports = {
   'mobile-wiki.js': {
     pattern: `${assetsFolder}/mobile-wiki-*.js`,
-    limit: '511KB',
+    limit: '514KB',
```
JoshuaRogan, line 8: Ah. Had to do the same 😢

JoshuaRogan (approved):
This code looks solid. Very flexible and sustainable approach 💯

PR #1944, opened 2019-11-25
IW | reduce worker count per pod and increase number of pods to havethe same total number of processes

rybmat:
## Description

Recently we can see that mobile-wiki containers are restarted. It may be due the fact that they operate at memory limit (see grafana dashboard). Lets reduce number of node-cluster workers and increase number of pods instead.

## Reviewers
@Wikia/iwing

### DIFF
--- a/fastboot-server/app/server.js
+++ b/fastboot-server/app/server.js
@@ -13,6 +13,7 @@ const server = new FastBootAppServer({
   afterMiddleware: middlewares.after,
   distPath: config.distPath,
   gzip: true,
+  workerCount: 16,
 });
 
 server.start();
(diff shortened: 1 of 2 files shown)


fastboot-server/app/server.js
```diff
@@ -13,6 +13,7 @@ const server = new FastBootAppServer({
   afterMiddleware: middlewares.after,
   distPath: config.distPath,
   gzip: true,
+  workerCount: 16,
```
rybmat, line 16: by default it is 32

nikodamn (approved):

mszabo-wikia (approved):

latata (approved):

PR #1945, opened 2019-11-25
Don't send request when id is null

JoshuaRogan:
## Description

Add null check to prevent unnecessary requests 

## Reviewers
@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots.js
+++ b/app/services/affiliate-slots.js
@@ -321,7 +321,7 @@ export default Service.extend({
       }
 
       // check if we have possible units (we can fail early if we don't)
-      if (!this._getAvailableUnits()) {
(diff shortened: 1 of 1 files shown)


mklucsarits (approved):

cschucode (commented):
How will this work on the Search Results page?

JoshuaRogan:
> How will this work on the Search Results page?

@cschucode  There are two functions `fetchUnitsForSearch` and `fetchUnitsForPage`. I would try to take a deep look at these files to fully understand them when you have a moment

cschucode:
Thanks and makes sense. While it is kind of obvious in the code, it goes back to @mklucsarits 's point about the difference between **Page** and **Article** and naming used.

cschucode (approved):

PR #1950, opened 2019-11-26
cake-5294 | Set up ddb affiliate units

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5294

## Description

Create DDB affiliate units in JSON file and add targeting info.

Note: 
Due to the increase in size of JSON files, bundle size was increase by 1KB.

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots-targeting.json
+++ b/app/services/affiliate-slots-targeting.json
@@ -41,5 +41,12 @@
     "wikiId": [
       673
     ]
+  },
+  {
+    "campaign": "ddb",
+    "category": "ddb",
+    "wikiId": [
+      51
+    ]
   }
 ]
--- a/app/services/affiliate-slots-targeting.json
+++ b/app/services/affiliate-slots-targeting.json
@@ -44,9 +44,6 @@
   },
   {
     "campaign": "ddb",
-    "category": "ddb",
-    "wikiId": [
-      51
-    ]
+    "category": "ddb"
   }
 ]
--- a/app/services/affiliate-slots-units.json
+++ b/app/services/affiliate-slots-units.json
@@ -220,8 +220,7 @@
     "heading": "The Official Digital Toolset for Dungeons & Dragons",
     "subheading": "Visit D&D Beyond",
     "launchOn": "2019-12-12T00:00:00Z",
-    "link": "https://www.dndbeyond.com/intro",
-    "links": null
+    "link": "https://www.dndbeyond.com/intro"
   },
   {
     "campaign": "ddb",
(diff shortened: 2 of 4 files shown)


app/services/affiliate-slots-units.json
```diff
@@ -211,5 +211,28 @@
...
+  {
+    "campaign": "ddb",
+    "category": "ddb",
+    "tagline": "Related Posts",
+    "image": "https://static.wikia.nocookie.net/0cebb500-f8c3-4f44-b4aa-f44e33fc2319",
+    "heading": "The Official Digital Toolset for Dungeons & Dragons",
+    "subheading": "Visit D&D Beyond",
+    "launchOn": "2019-12-12T00:00:00Z",
+    "link": "https://www.dndbeyond.com/intro",
+    "links": null
```
mklucsarits: Can we just leave this out? Or it has to be `null`?

mklucsarits (approved):

app/services/affiliate-slots-units.json
cschucode: Good call. 4f3dc48

app/services/affiliate-slots-targeting.json
```diff
@@ -41,5 +41,12 @@
     "wikiId": [
       673
     ]
+  },
+  {
+    "campaign": "ddb",
+    "category": "ddb",
+    "wikiId": [
+      51
```
vforge: What wiki is this?

app/services/affiliate-slots-units.json
```diff
@@ -211,5 +211,26 @@
...
     "isBig": true
+  },
+  {
+    "campaign": "ddb",
+    "category": "ddb",
+    "tagline": "Related Posts",
+    "image": "https://static.wikia.nocookie.net/0cebb500-f8c3-4f44-b4aa-f44e33fc2319",
+    "heading": "The Official Digital Toolset for Dungeons & Dragons",
+    "subheading": "Visit D&D Beyond",
+    "launchOn": "2019-12-12T00:00:00Z",
```
vforge: Is that correct? It's 12/11 4pm PDT.

cschucode: No. Definitely 12/12. Will get a specific time.

cschucode: From Keiko, going with 11am (PST) on Dec. 12, 2019.

app/services/affiliate-slots-targeting.json
cschucode: Just realized this was for filtering. Will remove because will be driven by Taxonomy ™️

vforge (approved):

PR #1970, opened 2019-12-06
CAKE-5363 | Sort recommendation level (page first)

JoshuaRogan:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5363

## Description

* Filter out community-level targeting on pages until we want to support that. 

## Reviewers
@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots.js
+++ b/app/services/affiliate-slots.js
@@ -95,8 +95,7 @@ const sortPageLevelFirst = (a, b) => {
     return 0;
   }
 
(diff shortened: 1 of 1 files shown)


vforge (approved):

vforge (approved):
I’d actually hardcode page vs community. No need to use string compare (or locale string compare) for that - this is string consts we’re using.

vforge (approved):

PR #2004, opened 2019-12-13
CAKE-5389 | Enable hulu campaign on specific communities

JoshuaRogan:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5389

## Description

* Show a big unit and small unit on article page when it's in the hulu wiki set
* Show a big unit on the search page when it's in the hulu wiki set
* Style units

## Reviewers

@Wikia/cake


vforge (commented):
LGTM

mklucsarits:
Looking good

vforge (approved):

JoshuaRogan:
The way the code is now it looks like there is no need for the big unit in the JSON file. It's always using the non big one

cschucode:
CAKE-5346 is asking for 3-item and Search units as well. This is going to require a small unit for subheading text as well as a change in the `_getHuluUnit` logic.

Unless, we want it to say `Recommended For You` on all Hulu units no matter where they are.

PR #2037, opened 2020-01-07
CAKE-5372 | Deep copy the units

JoshuaRogan:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5372

## Description

* There is a minor mutation to the link in the unit. Copying will resolve any conflicts between loads. 

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots.js
+++ b/app/services/affiliate-slots.js
@@ -227,7 +227,7 @@ export default Service.extend({
     currentTargeting.forEach((target) => {
       // we're checking all units
       availableUnits.forEach((unit) => {
-        if (unit.campaign === 'ddb' && unit.category === 'ddb') {
+        if (unit.campaign === target.campaign && unit.category === target.category) {
           const updatedUnit = this._updateUnitLink(unit, pageId);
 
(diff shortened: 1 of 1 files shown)


app/services/affiliate-slots.js
```diff
@@ -226,7 +227,7 @@ export default Service.extend({
     currentTargeting.forEach((target) => {
       // we're checking all units
       availableUnits.forEach((unit) => {
-        if (unit.campaign === target.campaign && unit.category === target.category) {
+        if (unit.campaign === 'ddb' && unit.category === 'ddb') {
```
JoshuaRogan: Oops need to remove

JoshuaRogan (commented):
.

cschucode (approved):
Tested on Dev. LGTM.

app/services/affiliate-slots.js
mklucsarits: Can you explain why DnDBeyond has a special case here?

JoshuaRogan: It's not, debug code. Thanks for catching

PR #2038, opened 2020-01-07
CAKE-5384 | Make affiliate unit classes cryptic

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5384

## Description
Make affiliate unit classes more cryptic so they won't be picked up by ad blockers.

## Reviewers

@Wikia/cake 
@mklucsarits


mklucsarits (approved):
That should work well enough.

PR #2046, opened 2020-01-09
CAKE-5395 | Support dark mode for big affiliate unit

JoshuaRogan:
## Links
* http://wikia-inc.atlassian.net/browse/CAKE-5395

## Description
* Add custom styles for the big affiliate unit in dark mode. 

## Reviewers
@Wikia/cake

### DIFF
--- a/app/styles/component/_affiliate-unit.scss
+++ b/app/styles/component/_affiliate-unit.scss
@@ -79,6 +79,10 @@
       margin: 0 -4px 16px;
     }
   }
+
+  .wds-button.wds-is-secondary {
+    cursor: pointer;
+  }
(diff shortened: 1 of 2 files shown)


cschucode (changes requested):
linting  - `npm run linter`

cschucode:
Looks like the related posts header is going to need the same treatment.

image

JoshuaRogan:
> Looks like the related posts header is going to need the same treatment.
> 
> image

Discussed offline. I am going to remove the gradient for now and discuss with derek if we want to add it to both of the modules.

cschucode (approved):

PR #2047, opened 2020-01-10
CAKE-5427 | Adjust Spacing & Gradient Color on Affiliate Units

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5427
* zpl://screen?sid=5db233296a01342c331fe781&pid=5c34f8570b5f05bdc9806cb9

## Description

There was too much space between affiliate units and ads beneath. This PR reduces the distance by using negative margin without negatively impacting the slant gradient.

Also, a bug was noticed that when using dark mode in the mobile app, the gradient needed to have a black rather than white background to stay consistent with the portable infobox background color.

## Reviewers

@Wikia/cake 
@JoshuaRogan

### DIFF
--- a/app/styles/component/_affiliate-unit.scss
+++ b/app/styles/component/_affiliate-unit.scss
@@ -90,7 +90,7 @@
   margin: 28px -.75rem -22px;
 
   &__slant {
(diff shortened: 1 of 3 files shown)


JoshuaRogan (approved):
Code LGTM . Nice description too!

PR #2048, opened 2020-01-11
fix-midlines | Use rgba values and color variable

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5395

## Description

Switched linear gradient colors back to **rgba** values. The errors might have only been on desktop and not actual android mobile device but because we can't test in live environment, this approach seems cleaner and worked when manually input into verify devtools console.

## Reviewers
@Wikia/cake 
@JoshuaRogan

### DIFF
--- a/app/styles/theme/_dark.scss
+++ b/app/styles/theme/_dark.scss
@@ -130,7 +130,7 @@ body {
 
 .aff-big-unit {
   &__header {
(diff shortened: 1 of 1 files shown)


JoshuaRogan:
Lgtm

PR #2050, opened 2020-01-12
IW-2389 | Variable rate limiting for Video Bridge

bkoval:
@Wikia/iwing


app/templates/head.hbs
```diff
@@ -167,10 +167,15 @@
 {{#if model.twitterDescription}}
   <meta name="twitter:description" content={{model.twitterDescription}}>
 {{/if}}
+{{! Twitter end }}
```
rybmat, line 176: do we need this comment?

app/templates/head.hbs
```diff
@@ -167,10 +167,15 @@
...
+
 {{#if model.isDedicatedForArticle}}
   <meta name="featured-video" content={{model.isDedicatedForArticle}}>
 {{/if}}
-{{! Twitter end }}
+
+{{#if model.playerImpressionsPerSession}}
+  <meta name="player-impressions-per-session" content={{model.playerImpressionsPerSession}}>
```
rybmat, line 184: Can you leave a comment why this needs to be in meta tag instead of shoebox?

app/templates/head.hbs
bkoval: It's an old comment, I just moved it to a right place. I assume it was added to group all the twitter-related meta tags together.

rybmat (approved):

PR #2061, opened 2020-01-17
Remove obsolete A/B tests

rybmat:
Why am I removing them?
1) they finished at least year ago
2) there is no product person in the company that would know what these experiments are about
3) As for now, the decision is that our internal abtesting tool will not be ported to ucp, so they would require changes anyway

@Wikia/iwing


mszabo-wikia (approved):
Can we reduce bundle size check value?

bkoval (approved):

app/components/recirculation-prefooter.js
```diff
@@ -86,27 +80,18 @@ export default Component.extend(
     actions: {
       postClick(post, index) {
         const labels = ['footer', `footer-slot-${index + 1}`];
-        let additionalRecommendationData;
 
-        if (this.shouldUseExperimentalRecommendationService) {
```
MikolajBalcerek, line 91: Seems like mobile wiki was never changed to permanently use the recommendations service, like in /app https://github.com/Wikia/app/pull/16934/files
Gonna have to change it in another ticket

MikolajBalcerek (approved):

PR #2100, opened 2020-02-14
IW-2956 | fix canPlayVideo method so it returns false when video is not present

xkxd:
http://wikia-inc.atlassian.net/browse/IW-2956

@Wikia/iwing @Wikia/adeng 

to check if video will be played on page you should call `window.canPlayVideo && window.canPlayVideo();`

### DIFF
--- a/vendor/inline-scripts/featured-video.js
+++ b/vendor/inline-scripts/featured-video.js
@@ -1,5 +1,5 @@
 (function () {
-  var canPlayVideoFlag = null;
+  var hasVideoOnPage = null;
 
   function getCookieValue(cookieName) {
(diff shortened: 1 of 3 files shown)


vendor/inline-scripts/featured-video.js
```diff
@@ -29,7 +29,13 @@
   }
 
   window.canPlayVideo = function () {
```
rybmat: is there a reason why we are not using ember service for such functionality?

rybmat (approved):

vendor/inline-scripts/featured-video.js
latata: It's because we want to use it outside ember

latata (approved):

fraszczakszymon (approved):

PR #2144, opened 2020-03-12
CAKE-5511 | Add proxy service

JoshuaRogan:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-5511

## Description
* Route all affiliate links through the redirect service

Instead of sending them to `disney.com/xkdhj39` let's send them to our own service which will redirect them: `https://services.fandom.com/affiliate/redirect/any/data/we/want?r=disney.com/xkdhj39`

## Reviewers

@Wikia/cake


app/utils/affiliate.js
```diff
@@ -0,0 +1,46 @@
...
+  const tracking = trackingPairsToObject(unit);
+
+  const category = unit.category;
+
+  // wikiId/articleId/category/algorithm/method/version
+  const path = `${wikiId}/${articleId}/${category}/${tracking.algo}/${tracking.method}/${tracking.version}`;
+
+  const potentialLink = `${host}${path}?r=${encodeURIComponent(link)}`;
+
+  // if we reach the limit lets serve them the link without the tracking
```
vforge, line 44: 👍

vforge: Can we hit the tracking with some specific value so at least we know how frequently this is triggered?

JoshuaRogan: Discussed offline and agreed that this event is very unlikely to happen (if ever).

JoshuaRogan: Discussed offline and agreed that this event is very unlikely to happen (if ever).

vforge (approved):

app/utils/affiliate.js
```diff
@@ -0,0 +1,50 @@
+import config from '../config/environment';
+
+function getServiceHost() {
+  if (config.environment === 'production') {
```
latata, line 4: you can use `fetch` service to get the env-specific service url (https://github.com/Wikia/ember-fandom/blob/master/addon/services/fetch.js). you will need to convert this util to the ember service

app/utils/affiliate.js
```diff
@@ -0,0 +1,50 @@
...
+    unit.tracking.forEach((kv) => {
+      tracking[`${kv.key}`] = kv.val;
+    });
+  }
+
+  return tracking;
+}
+
+export function linkToProxyLink(link, unit, wikiId, articleId = -1) {
+  // deploy to one community (marvel)
```
cschucode, line 29: I'm curious why marvel needs a link to proxy link?

vforge (approved):

app/utils/affiliate.js
JoshuaRogan: We are just doing a limited roll out. We will be doing this for every community.

PR #2150, opened 2020-03-20
IW-3175 | Mobile article edit entrypoints

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/WI-3175
* https://github.com/Wikia/unified-platform/pull/1560

## Reviewers

@Wikia/iwing

### DIFF
--- a/app/styles/component/wikia-ui-components/_wiki-page-header.scss
+++ b/app/styles/component/wikia-ui-components/_wiki-page-header.scss
@@ -30,14 +30,21 @@
   &__title-wrapper {
     margin-bottom: 1.875rem;
     width: 100%;
+
+    &.has-edit-link {
+      margin-bottom: 1.07rem;
+    }
   }
 
   &__edit-link {
     color: $wds-fandom-color-dark-gray;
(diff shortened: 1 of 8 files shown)


mszabo-wikia (approved):

app/styles/component/wikia-ui-components/_wiki-page-header.scss
```diff
@@ -30,5 +30,22 @@
   &__title-wrapper {
     margin-bottom: 1.875rem;
     width: 100%;
+
+    &.has-edit-link {
+      margin-bottom: 1.07rem;
```
latata, line 35: :o do we really need to use `rem` here?

latata (approved):

app/styles/component/wikia-ui-components/_wiki-page-header.scss
rybmat: It was using rem so I decided not to change it. It was discussed during mobile premiumisation project and rem was chosen, but why? I don't remember

PR #2158, opened 2020-03-26
IW-2854 | bring back recommendations from recommendations service

rybmat:
## Links

* http://wikia-inc.atlassian.net/browse/IW-2854

## Description

* bring back recommendations fetching from recommendations service and stop treating it as experiment

## Reviewers
@Wikia/iwing 
@Wikia/data-engineering

### DIFF
--- a/app/components/recirculation-prefooter.js
+++ b/app/components/recirculation-prefooter.js
@@ -5,7 +5,6 @@ import { reads, and } from '@ember/object/computed';
 import { computed, observer } from '@ember/object';
 import { run } from '@ember/runloop';
 import InViewportMixin from 'ember-in-viewport';
-import { inGroup } from '../modules/abtest';
 import { logError } from '../modules/event-logger';
 import { track, trackActions } from '../utils/track';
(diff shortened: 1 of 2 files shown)


app/components/recirculation-prefooter.js
```diff
@@ -24,12 +25,15 @@ export default Component.extend(
...
     sponsoredContent: service(),
+    runtimeConfig: service(),
 
     classNames: ['recirculation-prefooter'],
     classNameBindings: ['items:has-items'],
 
     listRendered: null,
+    requestId: null,
     topArticles: null,
+    fallbackItems: null,
```
mikulskibartosz, line 36: is it used for anything else? If no, maybe the variable could be renamed to "recommendationFallbackItems"?

mikulskibartosz (approved):

app/components/recirculation-prefooter.js
rybmat: nah, it's not java, lets keep names short ;) The recommendations prefix IMO does not bring any value here

PR #2177, opened 2020-04-09
IW-3341 | extract table styles

latata:
## Links

* https://wikia-inc.atlassian.net/browse/IW-3341

## Description

No changes in styles - just moving stuff around


app/styles/module/article/_table-cell.scss
```diff
@@ -0,0 +1,47 @@
+th,
```
rybmat, line 1: comment why it is extracted would be nice here

rybmat (approved):

PR #2185, opened 2020-04-17
CAKE-5427 | Reduce Extra Space Below Affiliate Unit

cschucode:
## Links

* [Jira](http://wikia-inc.atlassian.net/browse/CAKE-5427)
* [Zeplin](https://zpl.io/VqJnKx7)

## Description

The CSS dilemma here was how to create a diagonal separator that faded straight down. Our previous solution used linear-gradient that was able to slant the div by 5 degrees but wasn't able to fade straight down afterwards causing either a big gap between content or a faint line visible at the bottom of its container.

This solution uses two divs inside of a wrapper. The first div creates the 5 deg slant and fades from the background color of the page to the background color of the infobox. The second div fades straight down from the background color of the infobox back to the background color of the page.

Seems simple now. 😉 

## Reviewers

@Wikia/cake 
@unijuglr


JoshuaRogan (approved):
Code and approach LGTM. Great to see a good description too!

vforge (approved):

PR #2187, opened 2020-04-21
ADEN-9914 | Add IndexExchange Identity Library

bckatarzyna:


app/modules/ads/index.js
```diff
@@ -339,10 +339,12 @@ class Ads {
     }
 
     const { bidders } = window.Wikia.adBidders;
-    const { slotService } = window.Wikia.adEngine;
+    const { identityLibrary, slotService } = window.Wikia.adEngine;
 
     const inhibitors = [];
 
+    identityLibrary.call();
```
fraszczakszymon: What do you think about moving it to `triggerInitialLoadServices`? Do we need to trigger this call on each mobile-wiki pv?

fraszczakszymon (approved):

PR #2216, opened 2020-05-15
IW-3489 | install Article Comments package

xkxd:
## Links

* http://wikia-inc.atlassian.net/browse/IW-3489

### KEEP IN MIND IT WILL NOT BE MERGED TO DEV

### DIFF
--- a/app/components/article-comments.js
+++ b/app/components/article-comments.js
@@ -1,16 +1,20 @@
 import Component from '@ember/component';
-import { computed, observer } from '@ember/object';
-import { not } from '@ember/object/computed';
-import { scheduleOnce } from '@ember/runloop';
-import { inject as service } from '@ember/service';
+import {computed, get, observer} from '@ember/object';
+import {not} from '@ember/object/computed';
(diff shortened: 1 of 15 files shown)


app/components/article-comments.js
```diff
@@ -74,41 +71,17 @@ export default Component.extend({
...
     */
     toggleComments() {
-      const page = this.page;
-
       this.set('preserveScroll.preserveScrollPosition', true);
       this.toggleProperty('isCollapsed');
 
-      if (page !== null) {
+      if (this.isCollapsed) {
         this.set('page', null);
```
latata, line 110: do we still need page and pageCount propertY?

app/components/article-comments.js
```diff
@@ -122,42 +95,4 @@ export default Component.extend({
   scrollTop() {
     scrollToTop(this.element);
   },
-
-  fetchComments(page) {
```
rybmat: @xkxd does it mean we are not supporting comments from app once this change is released?

latata (approved):

PR #2220, opened 2020-05-19
Cake 5678 - Create new Campaign for HBO Max in Mobile and Desktop in-content Affiliate units

unijuglr:
## Links

* https://wikia-inc.atlassian.net/browse/CAKE-5678

## Description
Adding config and a slight template addition to support HBO MAX campaign
Can go live without concern, because it will be controlled by db rows being added at launch time

## Reviewers

@Wikia/cake

### DIFF
--- a/app/services/affiliate-slots-units.json
+++ b/app/services/affiliate-slots-units.json
@@ -7,7 +7,7 @@
     "subheading": "START YOUR FREE TRAIL",
     "link": "https://www.hbomax.com/gzm",
     "image": "https://static.wikia.nocookie.net/2ef28d1f-a27b-4ba3-943e-33948fddd089",
-    "extra_disclaimer": "Free trail for new customers only. Retrictions Apply."
+    "extra_disclaimer": "Free trail for new customers only. Restrictions Apply."
   },
   {
     "campaign": "ddb",
--- a/app/services/affiliate-slots-units.json
+++ b/app/services/affiliate-slots-units.json
@@ -4,20 +4,10 @@
     "category": "hbomax",
     "country": ["US"],
     "heading": "Where HBO Meets So Much More",
-    "subheading": "START YOUR FREE TRAIL",
+    "subheading": "START YOUR FREE TRIAL",
     "link": "https://www.hbomax.com/gzm",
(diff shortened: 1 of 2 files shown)


app/services/affiliate-slots-units.json
```diff
@@ -1,4 +1,24 @@
 [
+  {
+    "campaign": "hbomax",
+    "category": "hbomax",
+    "country": ["US"],
+    "heading": "Where HBO Meets So Much More",
+    "subheading": "START YOUR FREE TRAIL",
```
cschucode: TRAIL or TRIAL?

app/services/affiliate-slots-units.json
```diff
@@ -1,4 +1,24 @@
...
+    "link": "https://www.hbomax.com/gzm",
+    "image": "https://static.wikia.nocookie.net/2ef28d1f-a27b-4ba3-943e-33948fddd089",
+    "extra_disclaimer": "Free trail for new customers only. Restrictions Apply."
```
cschucode: (see above)

app/services/affiliate-slots-units.json
```diff
@@ -1,4 +1,24 @@
...
+  },
+  {
```
cschucode, line 12: Did this get added a second time?

app/services/affiliate-slots-units.json
```diff
@@ -1,4 +1,24 @@
...
+    "campaign": "ddb",
+    "category": "ddb-test-unit-id",
```
cschucode: Looks like we already have this unit.

cschucode (changes requested):
Requested:
* fix typos
* remove extra unit

app/services/affiliate-slots-units.json
unijuglr: weird...

cschucode:
Are we going to need a big unit as well? There should be two for every category/campaign. One with the "isBig" set to `true` and one without it for the `post-search-results` unit?

cschucode (approved):
LGTM.

PR #2292, opened 2020-07-17
Moving the scroll speed to only track to the DW, per nAndy

unijuglr:
## Links

* more relevant links

## Description
Just changing the parameter for where this routes, so that we can cut it off from flooding GA. This was turned off at the GTM level for 2 weeks previously, with no impact noticed. It only went back on because we had to pull jQuery out of the mix, and that code in GTM was reliant on it. 

Going forward, planning on updating code in the codebase to effect such changes, rather than override in GTM. This is just the first example. 

## Reviewers

### DIFF
--- a/app/utils/track.js
+++ b/app/utils/track.js
@@ -221,7 +221,7 @@ export function trackScrollY(time, scrollY) {
     action: 'scroll',
     category: 'scroll_speed',
     label: `${time}s|${scrollY}`,
-    trackingMethod: TrackingMethod.internal
+    trackingMethod: TrackingMethod.internal,
   });
 }
 
--- a/app/utils/track.js
+++ b/app/utils/track.js
@@ -145,7 +145,7 @@ export function track(params, usePrefix = true, force = false) {
   const isNonInteractive = params.isNonInteractive !== false;
   const pvUID = window.pvUID;
   const {
-    action, label = '', value = 0, trackingMethod = TrackingMethod.both,
+    action, label = '', value = 0, trackingMethod = TrackingMethod.internal,
   } = params;
 
   params = Object.assign({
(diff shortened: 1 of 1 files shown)


app/utils/track.js
```diff
@@ -145,7 +145,7 @@ export function track(params, usePrefix = true, force = false) {
   const isNonInteractive = params.isNonInteractive !== false;
   const pvUID = window.pvUID;
   const {
-    action, label = '', value = 0, trackingMethod = TrackingMethod.both,
+    action, label = '', value = 0, trackingMethod = TrackingMethod.internal,
```
nandy-andy: Oh, this looks wider I thought the change would be 🙈 It means all events on mobile-wiki will get to DW instead of DW&GA. I thought the change will be connected to `mercury-scroll_speed`. You mentioned you tested it with GTM, thought. When was the change?

nandy-andy (changes requested):
This change will move all events to DW instead of some of them. We can't do it because there is a need for `ads-babdetector-detection` event going to GA.

app/utils/track.js
nandy-andy: My proposal is to leave it as-is and pass `trackingMethod: TrackingMethod.internal` in https://github.com/Wikia/mobile-wiki/blob/6d7078cfd3e21c7cea82e2ceec5abb941513029f/app/utils/track.js#L220-L224

nandy-andy (approved):

app/utils/track.js
unijuglr: Oh, weird... I think I got my files confused. Guess that's why we do PRs, right. Thanks nAndy, I'll get it fixed over the weekend.

PR #2293, opened 2020-07-17
IW-3255 | article comments thread view

mortigen:
## Links

* https://wikia-inc.atlassian.net/browse/IW-3255

## Description

Support Article Comments thread view


latata (approved):

app/components/article-comments.js
```diff
@@ -68,6 +68,14 @@ export default Component.extend(InViewportMixin, {
...
 
+    const { urlThreadId } = this.articleComments.getUrlThreadParams();
+
+    if (this.isUcp && urlThreadId) {
+      this.articleComments.fetchCount(this.articleTitle, this.articleNamespace)
+      this.articleComments.load({ title: this.articleTitle, namespace: this.articleNamespace });
+      this.toggleProperty('isCollapsed');
+    }
+
     if (page !== null) {
```
xkxd: It seems there might be a chance we'll call `fetchCommentsBasedOnPlatform` twice, how bout `else if`?

app/components/article-comments.js
```diff
@@ -68,6 +68,14 @@ export default Component.extend(InViewportMixin, {
 
     this._super(...arguments);
 
+    const { urlThreadId } = this.articleComments.getUrlThreadParams();
+
+    if (this.isUcp && urlThreadId) {
+      this.articleComments.fetchCount(this.articleTitle, this.articleNamespace)
```
xkxd: it's already called in `didEnterViewport`, so it seems it'll be called twice

app/components/article-comments.js
```diff
@@ -68,6 +68,14 @@ export default Component.extend(InViewportMixin, {
...
+      this.articleComments.load({ title: this.articleTitle, namespace: this.articleNamespace });
+      this.toggleProperty('isCollapsed');
```
xkxd: Are we sure we want to 'toggle' this instead of setting `true/false` directly?

app/components/article-comments.js
mortigen: Sure

app/components/article-comments.js
mortigen: I will fetch this in `didInsertElement` then

app/components/article-comments.js
mortigen: You're right, I will set this to `false`

kasperekt (approved):

app/components/article-comments.js
```diff
@@ -68,7 +68,23 @@ export default Component.extend(InViewportMixin, {
 
     this._super(...arguments);
 
-    if (page !== null) {
+    const { urlThreadId } = this.articleComments.getUrlThreadParams();
+
+    if (this.isUcp) {
+      // to make sure we won't show cached value we have to fetch these comments on FE
+      this.articleComments
```
xkxd: Moving this here will is a little regression imo.

It'll be called on each page view even if user won't scroll to the comment module/bottom of the page. In `didEnterViewport` it was called only when scrolled to the page bottom which reduced the total amount of calls to the service.

So my proposal would be to leave in in `didEnterViewport` unless it causes different errors.

mortigen: I moved it to new `loadCommentCount` method so it will not fire twice anymore. It couldn't sit only in `didEnterViewport` because in some cases it wouldn't fire

latata (approved):

latata (approved):

kasperekt (approved):

mszabo-wikia (approved):

latata (approved):

PR #2302, opened 2020-07-31
IW-4000 | copy comment deep-link

latata:
https://wikia-inc.atlassian.net/browse/IW-4000

### DIFF
--- a/app/services/article-comments.js
+++ b/app/services/article-comments.js
@@ -2,6 +2,12 @@ import Service, { inject as service } from '@ember/service';
 import { ArticleCommentsFetchError } from '../utils/errors';
 import { track, trackActions } from '../utils/track';
 
+const bannerNotificationTypesMap = {
+  notify: 'message',
+  confirm: 'success',
+  warn: 'warning',
+  error: 'alert',
+};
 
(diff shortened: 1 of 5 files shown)


mortigen (approved):

app/services/article-comments.js
```diff
@@ -2,6 +2,12 @@ import Service, { inject as service } from '@ember/service';
 import { ArticleCommentsFetchError } from '../utils/errors';
 import { track, trackActions } from '../utils/track';
 
+const bannerNotificationTypesMap = {
```
latata, line 5: The DS notification types are different than the ones in UCP. We could either leave this mapping or unify the types.

app/services/article-comments.js
```diff
@@ -108,11 +115,20 @@ export default Service.extend({
         }));
       };
 
+      const notify = (message, type, timeout = 5000) => {
+        this.wdsBannerNotifications.addNotification({
```
latata, line 119: The notification does not appear sticky (to the top of the viewport). That should be fixed.

rybmat (approved):

xkxd (approved):

PR #2316, opened 2020-08-11
IW-4118 | fix scroll on article comments page

mortigen:
## Links

https://wikia-inc.atlassian.net/browse/IW-4118

## Problem

Clicking on category link caused scrolling to bottom.

## Solution

There is a condition for article-comments component that checks if `page` variable is not `null` to scroll to article comments. But in case the `page` is `undefined` it was also scrolling because there was no condition for that.

### DIFF
--- a/app/components/article-comments.js
+++ b/app/components/article-comments.js
@@ -68,7 +68,7 @@ export default Component.extend(InViewportMixin, {
 
     this._super(...arguments);
 
-    if (page !== null) {
+    if (page !== null && page !== undefined) {
       this.set('isCollapsed', false);
       this.fetchCommentsBasedOnPlatform(this.page);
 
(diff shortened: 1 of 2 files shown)


app/components/article-comments.js
```diff
@@ -68,7 +68,7 @@ export default Component.extend(InViewportMixin, {
 
     this._super(...arguments);
 
-    if (page !== null) {
+    if (page !== null && page !== undefined) {
```
mszabo-wikia, line 71: In an old branch of mine, I added a basic test for this component: https://github.com/Wikia/mobile-wiki/blob/ember-octane-update/tests/integration/components/article-comment-test.js.

Do you think it could be extended to cover this case?

kasperekt: This could be written as `page == undefined` as well I think, although I'm not 100% that it is better.

kasperekt (approved):

app/components/article-comments.js
mortigen: > In an old branch of mine, I added a basic test for this component: https://github.com/Wikia/mobile-wiki/blob/ember-octane-update/tests/integration/components/article-comment-test.js.
> 
> Do you think it could be extended to cover this case?

I will add these tests here and extend for this case

PR #2361, opened 2020-09-25
IW-4359 | Use DS version without licencing for wikia.org

bkoval:
@Wikia/iwing

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "compression": "1.7.4",
     "cors": "2.8.5",
     "deep-extend": "0.6.0",
-    "design-system": "git://github.com/Wikia/design-system.git#22.14.0",
+    "design-system": "git://github.com/Wikia/design-system.git#bdfc0e0ec6cb505f31d01120ea30d78091c57730",
     "ember-cli-dependency-lint": "1.1.3",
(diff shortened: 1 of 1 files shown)


latata (approved):

package.json
```diff
@@ -45,7 +45,7 @@
     "compression": "1.7.4",
     "cors": "2.8.5",
     "deep-extend": "0.6.0",
-    "design-system": "git://github.com/Wikia/design-system.git#22.14.0",
+    "design-system": "git://github.com/Wikia/design-system.git#bdfc0e0ec6cb505f31d01120ea30d78091c57730",
```
kasperekt: We should first deploy design-system with version number.

kasperekt (changes requested):

package.json
bkoval: So far the process was: we first test on the commit hash and if everything is OK then we release DS and update the version in the clients.

kasperekt: The process probably is the same :)

kasperekt (approved):

package.json
bkoval: Ah, now I get it - so you've approved the changes but wanted to be extra secure so that nobody merges the version with the commit hash after testing? Makes sense!

PR #2366, opened 2020-09-30
IW | Fix body-bottom script path

rybmat:
## Description
For the pv_tracking cookie, use the scriptPath from wikiVariables service instead of parsing articlePath.

## Reviewers
@Wikia/iwing 
@Wikia/data-engineering

### DIFF
--- a/app/components/fastboot-only/body-bottom.js
+++ b/app/components/fastboot-only/body-bottom.js
@@ -38,8 +38,8 @@ export default Component.extend({
       'language',
       'qualarooUrl',
       'isTestWiki',
+      'scriptPath',
     );
-    wikiVariables.scriptPath = this.wikiVariables.articlePath.replace(/\/$/, '');
 
     return JSON.stringify(Object.assign({
       cookieDomain,


mszabo-wikia (approved):

kasperekt (approved):

app/components/fastboot-only/body-bottom.js
```diff
@@ -38,8 +38,8 @@ export default Component.extend({
       'language',
       'qualarooUrl',
       'isTestWiki',
+      'scriptPath',
     );
-    wikiVariables.scriptPath = this.wikiVariables.articlePath.replace(/\/$/, '');
```
JackThePie, line 42: For some reason, this fix brakes incrementing pv_number on articles in a language different than 'en'. However, if it causes problems on production it can be merged and we will work on a fix ASAP.

JackThePie: Live QA confirmed that the way of testing was wrong. Fix works.

JackThePie (approved):

PR #2386, opened 2020-11-03
CAKE-6064 | Add Message Wall Notifications

cschucode:
## Links

* http://wikia-inc.atlassian.net/browse/CAKE-6064

## Description

Adding message wall notifications to mobile wiki

## Reviewers
@Wikia/cake 
@mklucsarits


app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,44 @@
+import Mixin from '@ember/object/mixin';
+
+export default Mixin.create({
+  /**
+  * Constructs a localized message wall post notification body
+  * @param {Ember.Object} model
+  * @returns {string}
+  */
+  getMessageWallPostBody(model) {
+    const firstReplierName = model.get('latestActors.0.name')
```
mklucsarits: This field will be empty for an anonymous user's post. In that case we should set it to "A Fandom user" (localized and hopefully existing already in the translation files).

app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,44 @@
...
+    const postTitle = this.postTitleMarkup;
+
+    // need this data !!
+    const isOwnWall = firstReplierName === '// check model.uri for username? in Message Wall uri';
+    const postRemoved = '// need check to see if post has been removed';
```
mklucsarits, line 15: Something I may have neglected to mention is that each of these mixins corresponds to a single notification type, and those types are listed here https://github.com/Wikia/pandora/blob/master/service/on-site-notifications/src/main/java/com/wikia/services/onsitenotifications/domain/notification/NotificationFactory.java#L32-L34
Each notification's `model` should have a `type` property corresponding to one in that list. That means "post removed" should have its own file handling the formatting for the `wall-post-removed-notification` so there's no trick to determining if the post was removed, that notification type will be dealt with separately.

app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,44 @@
...
+  * Constructs a localized message wall post notification body
+  * @param {Ember.Object} model
+  * @returns {string}
+  */
+  getMessageWallPostBody(model) {
+    const firstReplierName = model.get('latestActors.0.name')
+    const postTitle = this.postTitleMarkup;
+
+    // need this data !!
+    const isOwnWall = firstReplierName === '// check model.uri for username? in Message Wall uri';
```
mklucsarits: You'll want to reproduce this function here I think https://github.com/Wikia/unified-platform/blob/master/extensions/fandom/DesignSystem/scripts/DesignSystemGlobalNavigationOnSiteNotificationsTextFormatter.js#L192-L199, using the `model.get('uri')` as the parameter.

However! This method is broken for some usernames containing underscores, which is what CAKE-5920 is attempting to fix, but that hasn't been deployed yet. Newer notifications (after CAKE-5920 is deployed) will explicitly have the wall owner's username as a field in the notification model.

app/mixins/message-wall-post-notification.js
mklucsarits: My earlier comment on this was wrong. The notification model's `type` field doesn't correspond to the types in the Pandora service. However, there's a notification type `message-wall-post-removed` that should handle removed posts in a different file from this one.

app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,44 @@
...
+    const postRemoved = '// need check to see if post has been removed';
+
+    // check for is own wall
+    if (isOwnWall) {
+        // check for is post removed
+        if (postRemoved) {
```
mklucsarits: Move this stuff to another file for the `message-wall-post-removed` type. (message-wall-post-removed-notification.js)

app/mixins/message-wall-reply-notification.js
```diff
@@ -0,0 +1,48 @@
...
+  * @param {Ember.Object} model
+  * @returns {string}
+  */
+  getMessageWallReplyBody(model) {
+    const totalUniqueActors = model.get('totalUniqueActors');
+    const hasTwoOrMoreUsers = totalUniqueActors > 1;
+    const firstReplierName = model.get('latestActors.0.name');
+
+    // need this check 
+    const isOwnWall = firstReplierName === '// check model.uri for name in Message Wall uri';
```
mklucsarits: The wall owner isn't `latestActors.0`, that's the first person replying. Same goes for the original wall post (file above). We need to get the current logged in user's name, wherever that's stored in mobile-wiki, and compare it to the username in the wall URI.

app/mixins/message-wall-post-notification.js
cschucode: I believe we should be able to use `this.getTranslatedMessage('username-anonymous')`.

app/mixins/message-wall-post-notification.js
cschucode: Moved that function to `utils/messagewall.js`.

app/mixins/message-wall-post-notification.js
cschucode: Separated into it's own file `message-wall-post-removed-notification.js`.

app/mixins/message-wall-reply-notification.js
cschucode: Correct. It looks like we can use `this.usernameMarkup` to retrieve **username**.

app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default Mixin.create({
+  /**
+  * Constructs a localized message wall post notification body
+  * @param {Ember.Object} model
+  * @returns {string}
+  */
+  getMessageWallPostBody(model) {
+    const firstReplierName = model.get('latestActors.0.name') || this.getTranslatedMessage('username-anonymous');
+    const wallOwner = getMessageWallOwner(model.get('url'));
```
mklucsarits: I think the property is `uri` not `url`

cschucode: The only other example I found for this was in `wiki-pages.js`.

mklucsarits: The notification model is pulled from here: https://github.com/Wikia/design-system/blob/master/app/models/wds-on-site-notifications/wds-on-site-notification.js#L33-L44
The properties you use should match up with that.

JoshuaRogan:
@mklucsarits We also need to handle the new metadata property correct?

mklucsarits:
Yes, this code will now need to handle the metadata property but also different parameter names. That code has been merged to master in both DS and unified-platform. See the changes to UCP here https://github.com/Wikia/unified-platform/pull/3737/files
If no message output is coming from `getTranslatedMessage` it could be because the parameters need updating.

app/mixins/message-wall-post-notification.js
```diff
@@ -0,0 +1,33 @@
...
+import { getMessageWallOwner, getPossiblyAnonActorName } from '../utils/message-wall';
+
+export default Mixin.create({
+  /**
+  * Constructs a localized message wall post notification body
+  * @param {Ember.Object} model
+  * @returns {string}
+  */
+  getMessageWallPostBody(model) {
+    let wallOwner = model.get('metadata') && model.get('metadata.username');
```
mklucsarits, line 11: I think the model getter is safe for `undefined` and `null` so you only need `let wallOwner = model.get('metadata.username');` by itself.

app/utils/message-wall.js
```diff
@@ -0,0 +1,16 @@
...
+    return null;
+  }
+
+  return result[1];
+}
+
+export function getPossiblyAnonActorName(model) {
+  return model.get('latestActors[0]') && model.get('latestActors[0].name')
+    ? model.get('latestActors[0].name')
+    : 'A Fandom User';
```
mklucsarits, line 15: Wasn't there a localized string for this you were already using?

PR #2409, opened 2020-12-11
ADEN-10720 Set fixed tlb height on selected wikis

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-10720

### DIFF
--- a/app/modules/ads/slots.js
+++ b/app/modules/ads/slots.js
@@ -168,6 +168,7 @@ export const slots = {
         firstCall: true,
         adProduct: 'top_leaderboard',
         slotNameSuffix: '',
+        defaultClasses: [],
         bidderAlias: 'mobile_top_leaderboard',
         group: 'LB',
         options: {},
@@ -186,6 +187,7 @@ export const slots = {
         avoidConflictWith: '.ad-slot',
         bidderAlias: 'mobile_in_content',
         cheshireCatSlot: true,
+        defaultClasses: ['hide'],
         slotNameSuffix: '',
         group: 'MR',
         options: {},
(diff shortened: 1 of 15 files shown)


dmnsobczak (approved):

app/modules/ads/slots.js
```diff
@@ -462,6 +464,43 @@ export const slots = {
...
+          adSlot.addClass('wrapper-gap-disabled'); // shrink with animation
+        } else {
+          const id = scrollListener.addCallback(() => {
+            if (!universalAdPackage.isFanTakeoverLoaded()) {
+              adSlot.addClass('wrapper-gap-disabled'); // shrink with animation
+
+              scrollListener.removeCallback(id);
+            }
+          });
+        }
```
Loczi94: This part is completely unreadable for me. 🙈 
What do you think about creating separate methods like `shrinkWithouAnimation` and `shrinkWithoutAnimation` so there is no need to leave comments in code?
Also maybe we could move if statement from line 490 and make else if statement in 488?

app/templates/components/ad-slot.hbs
```diff
@@ -1,6 +1,6 @@
 <div
   id={{name}}
-  class="ad-slot {{adEngine3ClassName}} {{hiddenClassName}}"
+  class="ad-slot {{adEngine3ClassName}} {{defaultClasses}}"
```
Loczi94, line 3: Will it work with an array?

tests/integration/components/article-content-test.js
```diff
@@ -90,11 +90,6 @@ module('Integration | Component | article content', (hooks) => {
       }}{{/article-content}}`);
 
     assert.dom(topLeaderboardSelector).exists({ count: 1 });
-    assert.equal(
-      find(topLeaderboardSelector).previousSibling,
-      find('.wiki-page-header'),
-      'previous element is site header',
-    );
```
Loczi94: What's wrong with this test?

app/templates/components/ad-slot.hbs
fraszczakszymon: Nope. This is why there is join(' '): https://github.com/Wikia/mobile-wiki/pull/2409/files#diff-c158578aa157904c3f0096a4f978d60017581fbd5338bb9ebf9b06fd40b0d1d6R195

tests/integration/components/article-content-test.js
fraszczakszymon: Thanks for catching. It was removed at the time when top-leaderboard had additional wrapper. It's no longer the case. Assertion reverted.

Loczi94 (approved):

nandy-andy (approved):

PR #2410, opened 2020-12-11
ADEN-10742 | Logo Replacement 2.0

Loczi94:
## Links

https://wikia-inc.atlassian.net/browse/ADEN-10742

### DIFF
--- a/app/styles/module/_ads.scss
+++ b/app/styles/module/_ads.scss
@@ -205,15 +205,4 @@ $closeButtonHeight: 40px;
   }
 }
 
-.new-logo {
-  animation-duration: 3s;
-  animation-fill-mode: forwards;
-  animation-name: fade-in;
-  max-height: 55px;
-  opacity: 0;
-  padding-left: 12px;
-}
-
-.tracking-pixel {
-  display: none;
-}
+@include logo-replacement(55px, 12px);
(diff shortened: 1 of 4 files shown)


app/styles/module/_ads.scss
```diff
@@ -205,15 +205,29 @@ $closeButtonHeight: 40px;
...
 
-.new-logo {
-  animation-duration: 3s;
-  animation-fill-mode: forwards;
-  animation-name: fade-in;
-  max-height: 55px;
-  opacity: 0;
-  padding-left: 12px;
-}
+@mixin logo-replacement($max-height, $padding-left) {
```
nandy-andy, line 208: This mixin is different than the one created in ad-engine and I wonder if in the sake of consistency we shouldn't make them accept the same params 🤔 https://github.com/Wikia/ad-engine/pull/948/files#diff-20e570b6d6f5f0fe51f509457857db4705fb7388b2f8a38ded22154a6ddc039cR61

nandy-andy (approved):

app/styles/module/_ads.scss
nandy-andy: Discussed on a call. Let's leave it as-is.

PR #2415, opened 2020-12-22
ADEN-10753 Implement gap for incontent ads

fraszczakszymon:
## Links

* http://wikia-inc.atlassian.net/browse/ADEN-10753

### DIFF
--- a/app/modules/ads/slots.js
+++ b/app/modules/ads/slots.js
@@ -46,6 +46,7 @@ function injectAffiliateDisclaimer() {
     templateService,
   } = window.Wikia.adEngine;
 
+  // TODO: consider using defaultTemplates property
   slotService.on('affiliate_slot', AdSlot.STATUS_SUCCESS, () => {
     templateService.init('affiliateDisclaimer', slotService.get('affiliate_slot'));
   });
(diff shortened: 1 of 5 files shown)


app/styles/module/_ads.scss
```diff
@@ -91,6 +91,34 @@ $closeButtonHeight: 40px;
...
+    min-height: 250px;
+    transition: min-height .35s linear;
+
+    &.is-loading {
+      animation: is-loading 2s infinite;
+    }
+
+    &.wrapper-gap-disabled {
+      min-height: 0;
+    }
```
Loczi94: How about creating mixin?

fraszczakszymon: Good point 👍🏻

Loczi94 (approved):

app/modules/ads/slots.js
```diff
@@ -46,6 +46,7 @@ function injectAffiliateDisclaimer() {
     templateService,
   } = window.Wikia.adEngine;
 
+  // TODO: consider using defaultTemplates property
```
nandy-andy, line 49: Ticket it please - we'll forget about it if it's only in the code/GitHub

PR #2423, opened 2021-01-13
ADEN-10723 | Refactor no ads handling

Loczi94:
## Links

Ticket: https://wikia-inc.atlassian.net/browse/ADEN-10723

Introduced changes are required due to changes in UCP code: https://github.com/Wikia/unified-platform/pull/4450/files#r555093405


dmnsobczak (approved):

app/modules/ads/core/stable-ads.js
```diff
@@ -68,15 +68,13 @@ class StableAds {
...
     const reasonConditionMap = {
-      noexternals_querystring: isQueryParamActive(queryParams.noexternals),
-      noads_querystring: isQueryParamActive(queryParams.noads),
-      mobileapp_querystring: isQueryParamActive(queryParams['mobile-app']),
-      noads_pagetype: adsContext.opts.pageType === 'no_ads',
-      load_failed: this.hasLoadFailed,
+      mobileapp_querystring: isQueryParamActive(queryParams['mobile-app']) ? 'mobileapp_querystring' : false,
+      noads_reasons: adsContext.opts.noAdsReasons,
+      load_failed: this.hasLoadFailed ? 'load_failed' : false,
     };
```
fraszczakszymon, line 74: Is this ok that we have mixed three types here? string, string[] and boolean. 
For:
```
{
  foo: 'foo',
  bar: [ 'reason1', 'reason2' ],
  x: false
}
```
we will track: `off_foo,off_reason1,reason2`, am I right? Is this correct?
Also, I would go for `string || null` types instead of `string || boolean` combination.

Loczi94: Yes, you are right - the same way of tracking is currently on production.
About types - yep, probably `string || null` would be better.

Loczi94: Yes, you are right - the same way of tracking is currently on production.
About types - yep, probably `string || null` would be better.

fraszczakszymon (approved):

PR #2448, opened 2021-03-08
ADEN-10799 Start using ad-tag-manager tags from inlined data instead of additional call to the service

nandy-andy:
## Links
* https://wikia-inc.atlassian.net/browse/ADEN-10799
* https://github.com/Wikia/ad-engine/pull/981
* https://github.com/Wikia/unified-platform/pull/4860

## Description
Bump ad-engine and start using ad-tag-manager inlined data.

### DIFF
--- a/app/modules/ads/targeting.js
+++ b/app/modules/ads/targeting.js
@@ -222,6 +222,7 @@ export const targeting = {
       ref: getRefParam(),
       esrb: adsContext.targeting.esrbRating,
       geo: window.Wikia.adEngine.utils.geoService.getCountryCode() || 'none',
+      adTagManagerTags: adsContext.targeting.adTagManagerTags || {},
     };
 
     if (window.pvNumber) {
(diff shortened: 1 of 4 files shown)


Loczi94 (approved):

app/modules/ads/targeting.js
```diff
@@ -222,6 +222,7 @@ export const targeting = {
       ref: getRefParam(),
       esrb: adsContext.targeting.esrbRating,
       geo: window.Wikia.adEngine.utils.geoService.getCountryCode() || 'none',
+      adTagManagerTags: adsContext.targeting.adTagManagerTags || {},
```
Loczi94: When we call Permutive in ad-engine we read key-vals from context one by one: https://github.com/Wikia/ad-engine/blob/825c8685a3c069f5ebddbfc6ba8a9d43ef8b180a/src/ad-services/permutive/index.ts#L90-L113 thus here we need to read and assign ad-tags from adsContext.targeting.adTagManagerTags also one by one.

PR #2450, opened 2021-03-11
SITE-557 | FCP track page view experiment

latata:


app/utils/track.js
```diff
@@ -203,6 +203,7 @@ export function trackPageView(isInitialPageView, uaDimensions) {
     M.trackingQueue.push(window.trackQuantcastPageView);
     M.trackingQueue.push(window.trackComscorePageView);
     M.trackingQueue.push((isOptedIn) => {
+      window.fandomFCPtrackPageView(false);
```
latata: `false` to prevent increasing the page view number as it will be increased by regular page view event

vendor/inline-scripts/pageview-tracking-experiment.js
```diff
@@ -0,0 +1,104 @@
...
+      pv_unique_id: genUID(),
+      pv_number: pvNumberComputed,
+      pv_number_global: pvNumberGlobalComputed,
+      url: window.location.href,
+      c: headData.wikiVariables.id,
+      a: headData.trackingDimensions[21],
+      lc: headData.wikiVariables.language.content,
+      x: headData.wikiVariables.dbName,
+      s: 'mercury',
+      n: headData.trackingDimensions[25],
```
latata, line 50: namespace number took from tracking dimensions - that was the easiest way to get it 🤷‍♂️

vendor/inline-scripts/pageview-tracking-experiment.js
```diff
@@ -0,0 +1,104 @@
...
+    const params = {
+      beacon: beacon,
+      cb: new Date().valueOf(),
+      session_id: sessionId,
+      pv_unique_id: genUID(),
+      pv_number: pvNumberComputed,
+      pv_number_global: pvNumberGlobalComputed,
+      url: window.location.href,
+      c: headData.wikiVariables.id,
+      a: headData.trackingDimensions[21],
```
latata, line 46: article id took from tracking dimensions - that was the easiest way to get it 🤷‍♂️

vendor/inline-scripts/pageview-tracking-experiment.js
```diff
@@ -0,0 +1,104 @@
...
+      const r = Math.random() * 16 | 0, v = c === 'x' ? r : (r & 0x3 | 0x8);
+
+      return v.toString(16);
+    });
+  }
+
+  function getEventParams(shouldIncreasePvNumber) {
+    let pvNumberComputed = pvNumber ? parseInt(pvNumber, 10) : 0;
+    let pvNumberGlobalComputed = pvNumberGlobal ? parseInt(pvNumberGlobal, 10) : 0;
+    if (shouldIncreasePvNumber) {
```
latata: for initial page view `shouldIncreasePvNumber` is true as the FCP tracking event is sent before the regular page view event. For the next SPA page views the event is run after the regular page view event so the value is already increased.

app/utils/track.js
```diff
@@ -203,6 +203,9 @@ export function trackPageView(isInitialPageView, uaDimensions) {
     M.trackingQueue.push(window.trackQuantcastPageView);
     M.trackingQueue.push(window.trackComscorePageView);
     M.trackingQueue.push((isOptedIn) => {
```
bkoval, line 205: Why do we need to wait for the tracking queue? Maybe we could remove everything that requires user opt in from that request and avoid depending on tracking opt-in load time?

Otherwise the event will still depend on additional factors and not only on the page load.

vendor/inline-scripts/pageview-tracking-experiment.js
bkoval: In theory, if readData() goes into catch block, head Data will be empty so this code will throw an error. Not a big deal for the experiment but when productizing this we'd probably want to have a fallback

app/utils/track.js
latata: it's only for subsequent page views so it won't depend on any other factors. The queue should be flushed immediately.

latata: https://github.com/Wikia/mercury-shared/blob/master/src/tracking/tracking-queue.js#L9

vendor/inline-scripts/pageview-tracking-experiment.js
latata: right, we can improve it later on

bkoval (approved):

PR #2454, opened 2021-03-19
SITE-670 | set CPU limits in k8s descriptors

rybmat:
## Links
* http://wikia-inc.atlassian.net/browse/SITE-670

## Description
Similar issue as in this RCA: https://wikia-inc.atlassian.net/wiki/spaces/SE/pages/1734443093/Increased+5xx+error+rate+in+discussion+service

For some reason, default 1.0 CPU limit started to be respected, therefore pods were throttled and that caused failures

## Reviewers
@Wikia/site-experience

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     ]
   },
   "fastbootDependencies": [
+    "abortcontroller-polyfill",
     "bunyan",
     "crypto",
     "deep-extend",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -145,6 +145,7 @@
     ]
   },
   "fastbootDependencies": [
+    "abortcontroller-polyfill",
```
rybmat, line 148: solves `Package 'abortcontroller-polyfill' is required to be in the whitelist.` warnings

lomiarz (approved):

mortigen (approved):

PR #2468, opened 2021-03-25
PLATFORM-5761 Set global banner for Wiki Rules and Blocking Policy notice

mmeller-wikia:
https://wikia-inc.atlassian.net/browse/PLATFORM-5761

## Description

## Reviewers

### DIFF
--- a/app/services/migration-notification.js
+++ b/app/services/migration-notification.js
@@ -1,5 +1,6 @@
 import Service, { inject as service } from '@ember/service';
 import localStorageConnector from '@wikia/ember-fandom/utils/local-storage-connector';
+import Cookies from 'js-cookie';
 
 export default Service.extend({
   fastboot: service(),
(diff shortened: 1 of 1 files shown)


app/services/migration-notification.js
```diff
@@ -72,9 +73,7 @@ export default Service.extend({
 
   shouldShowWikiRulesAndBlockingPolicyBanner() {
     return (this.wikiVariables.wikiRulesBlockingPolicyBanner
-      && localStorageConnector.getItem(
-        this.wikiRulesBlockingPolicyStorageKey,
-      ) !== this.storageTrueValue);
+            && Cookies.get(this.wikiRulesBlockingPolicyStorageKey) !== this.storageTrueValue);
```
abador, line 76: please add comments that this is only for this case

abador (approved):

app/services/migration-notification.js
t-tomalak, line 76: I think we should leave previous rule, because when we will release this new version users which already dismissed this message in local storage will see that again.

t-tomalak (changes requested):

t-tomalak (approved):
I gues if it works it works 🙂

